Issue #28, opened 2017-10-14
floating point literals


luciusmagn:
Parse floating point numbers. Rhai already supports them, but has no way of actually creating them:
```rust
let x = 5.00001;
```

rustysec:
submitted #38 for this using this syntax
```js
let x = f5_00001;
```

rustysec:
based on @luciusmagn feedback, updated the syntax to 
```rust
let x = 5.0001;
```

rustysec:
can probably close per #38

luciusmagn:
Ah, yes

PR #38, opened 2017-10-28
candidate f64 support

rustysec:
I ran into a situation where I needed floating point support and ended up doing some hacking around with string passing and converting outside of the Rhai script. Then I was looking through the issues and noticed #28, thought I'd take a crack at it. Basically this short circuits the token parsing to add an additional literal type that starts with "f" and attempts to treat it like a float (f64 only at this point). There are probably some more efficient or elegant ways to do this, but I didn't want to mess with the rest of the parsing engine unless this solution is not acceptable.

an example of usage would be
```
let x = f1_0;
let y = f1_1;
print(x < y)
```
which will yield true. i added some test cases around the floats and all others are passing as well. i'm completely open to re-implementing this a different way.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rhai"
-version = "0.5.1"
+version = "0.6.2"
 authors = ["Jonathan Turner", "Lukáš Hozda"]
 description = "Embedded scripting for Rust"
 homepage = "https://github.com/jonathandturner/rhai"
--- a/README.md
+++ b/README.md
@@ -37,10 +37,34 @@ The repository contains several examples in the `examples` folder:
 - `reuse_scope` evaluates two pieces of code in separate runs, but using a common scope
 - `rhai_runner` runs each filename passed to it as a Rhai script
 - `simple_fn` shows how to register a Rust function to a Rhai engine
+- `repl` a simple REPL, see source code for what it can do at the moment
 
 Examples can be run with the following command:
 ```bash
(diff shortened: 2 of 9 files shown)


rustysec:
This does have the downside of shadowing valid variable names.
```rust
let f1 = 10;
let x = f1;
print(x)
```
will print `10` instead of `1` like you might expect.

luciusmagn:
hmm, apart from the syntax, I like the changes. Maybe we could look into parsing floats with a familiar syntax? 

Line 273 in [parser.rs](https://github.com/jonathandturner/rhai/blob/master/src/parser.rs#L273) is where we parse (at the moment) integer literals. Since our TokenIterator can do peeking, maybe we could peek if there is a dot after the last digit and if yes, try parsing digits behind it.

Let's ask @jonathandturner for his opinion on this, too.

rustysec:
I agree, the more I've used it the more I'm convinced I need to rework the syntax. I'll submit a different candidate tonight or tomorrow.

rustysec:
Pushed a more sane implementation with very standard syntax and without the token name collision I mentioned above.
```rust
let x = 1.99;
let y = 2.00;
print(x < y);
```
will produced the desired `true`.

luciusmagn:
This is exactly what I had in mind. Could you try resolving the conflicts in src/tests.rs created after I merged your stringly += PR? I could merge, then

luciusmagn:
Lovely, merging!

Issue #26, opened 2017-10-14
comments


luciusmagn:
Implement comments, possibly through very basic preprocessor, rather than by editing the parser. Through thorough discussion, we have settled on Rust/C/C++ style comments:
```rust
// I am a single-line comment
/* I am a multiline
   comment with multiple lines
   and a  /* nested comment */
*/
```

luciusmagn:
#41 provides a comment implementation that 100% follows the spec, therefore closing

PR #39, opened 2017-10-30
candidate comment implementation

rustysec:
I inlined the comment detection and parsing, which gets us almost to the spec in #26, however without nested comments. It also seems that implementing a preprocessor would be the best way to handle nesting and also maintain the performance profile of the engine.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -353,7 +353,37 @@ impl<'a> Iterator for TokenIterator<'a> {
                 '+' => return Some(Token::Plus),
                 '-' => return Some(Token::Minus),
                 '*' => return Some(Token::Multiply),
-                '/' => return Some(Token::Divide),
+                '/' => {
+                    if let Some(&comment) = self.char_stream.peek() {
+                        match comment {
+                            '/' => {
+                                self.char_stream.next();
+                                while let Some(&comment_next) = self.char_stream.peek() {
+                                    match comment_next {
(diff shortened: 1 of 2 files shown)


luciusmagn:
Hmm, I have a slightly different implementation, which handles nesting in a very simple way:
```rust
       '/' => {
                    match self.char_stream.peek() {
                        Some(&'/') => {
                            self.char_stream.next();
                            while let Some(c) = self.char_stream.next() {
                                if c == '\n' { break; }
                            }
                        }
                        Some(&'*') => {
                            let mut level = 1;
                            self.char_stream.next();
                            while let Some(c) = self.char_stream.next() {
                                match c {
                                    '/' => if let Some('*') = self.char_stream.next() {
                                        level+=1;
                                    }
                                    '*' => if let Some('/') = self.char_stream.next() {
                                        level-=1;
                                    }
                                    _ => (),
                                }

                                if level == 0 {
                                    break;
                                }
                            }
                        }
                        _ => return Some(Token::Divide),
                    }
                }
```

Thanks a lot for the effort, but in this case, I will be merging my solution as it fulfills the spec 100%

rustysec:
Awesome, I actually arrived at that same solution about an hour after I submitted the pr, hah! I'm still of the opinion that a pre-processor would be a good idea to handle scrubbing the comments from the code before execution. That's probably a post-1.0 task, however.

PR #42, opened 2017-10-31
candidate compound assignment implementation

rustysec:
First hack at this, I would like a second set of eyes on the operation precedence. Validated the tests cases against the python repl.


luciusmagn:
I think it's good, thank you for another contribution

Issue #24, opened 2017-10-14
binary operators


luciusmagn:
Implement the missing binary operators:
```rust
let rem = 10 % 4; // 2
let lsh = 10 << 4; // 160
let rsh = 10 >> 4; // 0
let and = 10 & 4; // 0
let or = 10 | 4; // 14
let xor = 10 ^ 4; // 14
```

luciusmagn:
#42 implements all but modulo

luciusmagn:
implemented in #47

Issue #44, opened 2017-10-31
Implement modulo


luciusmagn:
```rust
let x = 10 % 4; // == 2
```

Will close #24, too

luciusmagn:
implemented

PR #47, opened 2017-10-31
bugfix for binary and, modulo, binary ops tests

rustysec:
I found a bug in the binary & operator that I missed in test cases. Also, I saw #44 closed but couldn't find it in the code so I threw an implementation in here on a whim. I can separate these two out if you want cleaner commit tracking.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -7,7 +7,7 @@ use std::fmt;
 use parser::{lex, parse, Expr, Stmt, FnDef};
 use fn_register::FnRegister;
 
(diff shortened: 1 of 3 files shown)


rustysec:
To clarify, this should resolve #24.

luciusmagn:
Ah, awesome, another bug I discovered is that these operators aren't added to Token::is_next_unary(), so technically
```rust
let x = 5;
x += -5;
```
is invalid. I will add them ASAP

rustysec:
Dang it, sorry about that.

PR #50, opened 2017-11-03
separated tests for mergability

rustysec:
I kept some of the related tests grouped together, but we can obviously separate them further if that is desired. This should improve merging multiple branches in the future.


luciusmagn:
Awesome, thank you for your contribution. I don't think that we need to split the tests further, this seems to be good enough, at least for now.

PR #54, opened 2017-11-24
i64 and f64 exponent implementation

rustysec:
implemented 
```f64 ~ f64```
```f64 ~ i64```
```i64 ~ i64```
as well as a ```~=``` operator and some tests.


luciusmagn:
Thanks a lot! I had no idea Rust would support non-integer exponents, awesome :D

PR #55, opened 2017-11-25
Modules

luciusmagn:
Preparation pull request, not intended for merging yet


torkleyy:
Reviewed 7 of 18 files at r1.
Review status: 7 of 18 files reviewed at latest revision, 2 unresolved discussions.

---

*[Cargo.toml, line 19 at r1](https://reviewable.io:443/reviews/jonathandturner/rhai/55#-L0PdNCsFAByympojM05:-L0PdNCsFAByympojM06:b-8gejwt) ([raw file](https://github.com/jonathandturner/rhai/blob/293928e3e0a60678eff943e463e5d5f8fc5cf8a8/Cargo.toml#L19)):*
> ```TOML
> [features]
> default = ["modules"]
> modules = []
> ```

Why are you adding a new feature? Does this impose any additional overhead?

---

*[examples/repl.rs, line 31 at r1](https://reviewable.io:443/reviews/jonathandturner/rhai/55#-L0PdsiYTvgRoM_uGlY1:-L0PdsiYTvgRoM_uGlY2:biyf869) ([raw file](https://github.com/jonathandturner/rhai/blob/293928e3e0a60678eff943e463e5d5f8fc5cf8a8/examples/repl.rs#L31)):*
> ```Rust
> 
>     register(&mut engine);
>     engine.module_fns(register);
> ```

This looks a bit odd to me. Why would you want to register different functions for your modules?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/jonathandturner/rhai/55)*

luciusmagn:
1. It does add some overhead, yes. Slows down function calls and variable accessing by a small bit and increases compile times noticeably, but not drastically. The future idea of Rhai is to allow the programmer to opt-in/out to "pay" for just the features they need.

2. the register fn is a function which registers stuff to all modules. In that place, I do the first line to register them to the parent engine as well. The register fn is one of the things I am not decided yet about and I am considering other solutions of doing that.
As far as 'why would I want to register different functions to modules' goes, the answer is that I wouldn't I admit. The register function can do more than just registering functions, though. You can also register variables, execute extra Rhai code and perhaps call the Engine's functions.

torkleyy:
Thank you for the response.

Okay, paying only for the things you need sounds good.

Am I correct in assuming that rhai always executes function (or methods, which translate to a function where the receiving object is the first parameter) and has no notion of a field? If so, is that the reason why modules are using `::` over `.`? This is probably a rather stupid question, but I didn't find any sort of documentation for the language (and didn't have the time to fully read the source yet).

luciusmagn:
Rhai always executes the source code in modules, but never the functions
unless they are used within the module. The only reason modules are using
'::' instead of '.' is because Rust is using '::' instead of '.'. That's
all.

As far as documentation goes, there is just the README.md and
https://docs.rs/rhai/0.7.1/rhai/ . Documentation is still a major WIP.

2017-12-15 22:20 GMT+01:00 Thomas Schaller <notifications@github.com>:

schungx:
PR https://github.com/jonathandturner/rhai/pull/150 adds modules.

Issue #56, opened 2017-12-20
Question regarding the implementation


torkleyy:
In this line:

https://github.com/jonathandturner/rhai/blob/master/src/engine.rs#L735

Rhai sets `val` to `target`, which however is just `val` cloned. So I don't quite understand what that is doing. @luciusmagn Would you be so kind and explain to me what I'm missing and why the code is structured like this?

Issue #46, opened 2017-10-31
Easy interface for calling Rhai functions from Rust


luciusmagn:
Calling Rust functions from Rhai is simple and easy, the other way around, not so much. The way I see it, we need a macro or two. Therefore, we need to:

1. Make `call_fn()` public (Already done)
2. Make a macro like call!() for calling Rhai functions
3. Document everything

However, this open to discussion, as there might be better ways to handle function calling from Rhai

schungx:
The `call_fn` method is simplified in https://github.com/jonathandturner/rhai/pull/106

PR #57, opened 2017-12-20
Rewrite Rhai function dispatching

### LINKED ISSUE #51: More than six arguments?
Having read how Rhai functions are dispatched, I don't quite get why there is a fixed number of arguments limit. Arguments are basically passed as a six-elements-long homogeneous tuple of `Option<&mut Box<Any>>` only to assume last N of them are `None`. Not only does it introduce a argument number limit but implies quite a lot of code repetition. Wouldn't it be better to use slices instead of tuples?

Of course `fn_register` needs heterogeneousity for native Rust functions, at least until variadic generics come into play. But as for now, why not introduce some macro to automate N-arguments-precisely problem? Then it would be trivial to write another macro to abstract it away just to the limit of N. And even then, any limit doesn't seem really rational for Rhai-defined functions as it doesn't depend on variadic generics (it's dynamic anyway).

torkleyy:
~~I'm not sure about all the source I changed, I already created #56 because I was unsure what `get_dot_val` is doing in that line.~~

There are still some things unresolved:

* ~~Consider using `SmallVec` (external dependency, needs discussion given that no dependencies is listed as a feature)~~ [POSTPONED]
* [x] Replace `&mut Box<Any>` with `&mut Any`
* [x] Provide a nicer function calling interface (#46 #7)
* [x] Rewrite the rest of the procedural code as functional code
* [x] Properly test new function dispatching (I think there are some mistakes I made)

---

Fixes #51

### DIFF
--- a/examples/arrays_and_structs.rs
+++ b/examples/arrays_and_structs.rs
@@ -1,5 +1,5 @@
 extern crate rhai;
-use rhai::{Any, Engine, EvalAltResult, RegisterFn};
+use rhai::{Engine, RegisterFn};
 
 #[derive(Clone, Debug)]
 struct TestStruct {
@@ -16,22 +16,13 @@ impl TestStruct {
     }
 }
 
-fn print_ty_name(test: Vec<&mut Box<Any>>) -> Result<Box<Any>, EvalAltResult> {
-    println!("Ty name: {:?}", (*test[0]).as_ref().type_name());
-
-    Ok(Box::new(()))
-}
-
 fn main() {
     let mut engine = Engine::new();
 
     engine.register_type::<TestStruct>();
 
-    print_ty_name(vec![&mut (Box::new(TestStruct::new()) as Box<Any>)]).unwrap();
-
     engine.register_fn("update", TestStruct::update);
     engine.register_fn("new_ts", TestStruct::new);
-    engine.register_fn_raw("ty_name".to_owned(), None, Box::new(print_ty_name));
 
     println!(
         "{:?}",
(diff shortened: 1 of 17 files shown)


luciusmagn:
Howdy, thank you for the contribution, I see you did a lot of work. Tests still need fixing, but I can help with that. Once the tests are fixed for the current API, we will see if you missed anything. I will also review the code myself ASAP.

As far as `smallvec` goes, it is a great crate. If the API of SmallVec is 1:1 with Vec, we could maybe lock it behind a feature.

For the nicer interface, I recently discovered the possibility of creating functors (C++-ish functors as in stateful functions/callable objects, not what we mean by the FP definition) using the Fn, FnOnce and FnMut traits. The problem is that it requires two nightly features, so I am considering solutions for that.

I will tell you more once I have reviewed, compiled & tested the new stuff.

torkleyy:
Thank you for the fast response! Please note that this is still a work in progress, so yeah not everything compiles yet. While editing the code, I noticed some more things that could be improved, but I assume you prefer another PR for that so this one doesn't get too big.

As

luciusmagn:
It looks really good, I only have a few little exceptions to formatting, but that's probably the fault of rustfmt and is not a problem at all.

Rhai is not that big of a project yet (and hopefully it will never be a huge project), so with PRs' organization, you are at liberty.

I have reviewed the code, I like it and don't see anything suspicious (design-wise). I believe that the changes will simplify future development, not to mention that it allows me to remodel my modules in a more frugal way.

I will merge this one, and I'd be grateful if some of your other ideas came into being as well.

PR #61, opened 2017-12-31
Remove debug statements

thomas-jeepe:
It gets pretty annoying when trying to debug my own code


luciusmagn:
Ah yes, we can definitely turn them into some configurable macros, maybe lock it behind a feature

torkleyy:
I was talkin about logging macros provided by the `log` crate, actually.

luciusmagn:
temporarily resolved by a debug macro, will make an issue to continue this

PR #67, opened 2018-03-26
allow to reuse scope when calling a function. fixes #66

### LINKED ISSUE #66: variable not found in scope
I evaluate this script :
```
let angle=0.0;
fn update() {
    angle += 0.1;
}
```

Then when I try to evaluate `"update()"` using the same scope, I've got a `ErrorVariableNotFound("angle")` even though I see the variable in my scope :
```
[("angle", Any)]
```

Is it not supposed to work this way ?

jice-nospam:


schungx:
Not sure if this will be a particularly good idea...  although the implementation is simple enough.

With nested function calls, it will be very easy to forget which scope you're capturing a variable from and all the shadowing mess that goes with it...

~~~rust
let x = 42;

fn change_x() {
    x += 123;
}

fn foo() {
    let x = 0;
    change_x();
    print(x);
}
~~~

jice-nospam:
Ok I understand, I'm closing this PR

Issue #45, opened 2017-10-31
Return values from loops


luciusmagn:
A while ago, Rust introduced yet another amazing feature:
```rust
let x = loop {
    // complicated computations
    break 7;
};

assert_eq!(x, 7);
```

Would be cool if Rhai had it too. Maybe even for while & for (when it comes around)

rustysec:
just an update, i'm extremely close to getting this working 100%

luciusmagn:
Awesome! I am very close to working modules

2017-12-01 3:17 GMT+01:00 Russ <notifications@github.com>:

rustysec:
so this works, but without introducing a `()` type into the script engine we can end up in some weird states. a loop {} may not always return something, there's no enforcement of consistency at this point. for instance a while loop like this:
```
let i = 0;
while i < 100 {
    if i == 50 { break true; }
    i+=<user input>;
}
```
depends on the user input to increment. if it never hits 50, the loop returns with () at the end which the engine doesn't really know how to gracefully handle. something like this is a simpler example:
```
> let x = while true == false { break 7 }
> print(x)
error: Function argument types do not match
```
while this works as expected:
```
> let x = while true == true { break 7 }
> print(x)
7
```
 If anyone has ideas _other_ than me implementing a () type, i'd love to hear them!

schungx:
I'm wondering whether this is really necessary, as it is just as easy to write:

```
let x;

loop {
    // complicated computations
    x = 7;
    break;
};
```

This way, statements such as `while`, `loop`, `for` etc. do not need to be turned into valid expressions, avoiding some weird-looking scripts like `let x = while true == true { break 7 };`.

schungx:
I'm closing this one unless that are new opinions.

PR #70, opened 2018-05-23
added unit type as ()

rustysec:
very simply added the ability to parse an empty type as `()`, it's expressed internally as `Unit`. This is basically required for #45 to work correctly. if we want some keyword(s) like `unit`, `null`, `void`, etc to be attached to the type, i can add those in as well.

i also added a few test cases, although they don't provide full coverage of how this type might appear in the wild.

### DIFF
--- a/src/tests/unit.rs
+++ b/src/tests/unit.rs
@@ -1,0 +1,34 @@
+use engine::Engine;
+
+#[test]
+fn test_unit() {
+    let mut engine = Engine::new();
+
+    if let Ok(result) = engine.eval::<()>("let x = (); x") {
+        assert_eq!(result, ());
+    } else {
+        assert!(false);
+    }
+}
+
+#[test]
+fn test_unit_eq() {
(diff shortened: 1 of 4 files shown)


src/tests/unit.rs
```diff
@@ -0,0 +1,34 @@
...
+        assert!(false);
+    }
+}
+
+#[test]
+fn test_unit_eq() {
+    let mut engine = Engine::new();
+
+    if let Ok(result) = engine.eval::<bool>("let x = (); let y = (); x == y") {
+        assert!(result);
```
torkleyy: This should simply be `assert_eq!(engine.eval::<..>(..), Ok(true))`. The same is true for all the other tests, actually.

luciusmagn: Fair point, it is just that the first tests were written that way and the new ones followed, I might refactor them later, when I have time. Sadly, I don't have much of it this month

PR #71, opened 2018-06-13
Use default rust directory structure for tests

juanibiapina:


luciusmagn:
Looks good, thanks! No we have one historical baggage less.

PR #76, opened 2018-07-16
better not-found error messages for variables and functions

stevedonovan:
I've chosen to call the basic Rhai types 'integer', 'float', 'string' and 'array'.

Next step is providing API to register a type with its name and keep a type-name map. Then these error messages will be aware of custom user types.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -14,7 +14,7 @@ use call::FunArgs;
 
 #[derive(Debug)]
 pub enum EvalAltResult {
-    ErrorFunctionNotFound,
+    ErrorFunctionNotFound(String),
     ErrorFunctionArgMismatch,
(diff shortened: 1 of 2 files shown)


luciusmagn:
Awesome, thank you for this improvement. Rust actually has some sort of API for storing information about types, but it is still nightly. `TypeId` if I recall correctly

luciusmagn:
Yeah, and we could perhaps provide another macro, so that users don't necessarily have to call that function manually, something like:

```rust
reg_type!(engine, MyType);
```

stevedonovan:
I did think of a macro, but thought people would think it less than elegant :)

The story with TypeId is that it's basically just a large unique integer, so we have to provide the names ourselves. Rust does not have much runtime introspection by design.

PR #81, opened 2019-04-15
Fixed compile error due to type_id issues - stabilization through the

RektInPieces:
power of guess and check


RektInPieces:
https://github.com/rust-lang/rust/issues/58922

Is where the issue is reference

RektInPieces:
Yeah I think the other one fixes this as well so I'm just gonna go ahead and close this

PR #82, opened 2019-04-17
Fix ambiguous calls to `type_id()` preventing compilation on 1.35-nightly.

dkushner:
Just ran into this after compiling a project on nightly that integrates Rhai.


RektInPieces:
Yeah we found the same thing, but this is a superior way of doing it.

jonathandturner:
Thanks!

PR #86, opened 2019-09-16
Fix: README and a couple of mis-named types?

timfish:
Rhai looks great and hopefully I can contribute some more features, but for now this is all I have:

- It looks like `No unsafe code` [is no longer true](https://github.com/jonathandturner/rhai/blob/7d3306f6c7adb310897859bf03faca3f878a80a8/src/any.rs#L87)
- [This comment](https://github.com/jonathandturner/rhai/pull/77#issuecomment-457901347) pointed out a mis-named `f32` in that PR
- Then I noticed that `u64` was declared twice with the names `u64` and `usize`

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,6 @@ Rhai's current feature set:
 * Easy-to-use language based on JS+Rust
 * Support for overloaded functions
 * No additional dependencies
-* No unsafe code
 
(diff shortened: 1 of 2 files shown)


luciusmagn:
Hey, thanks for the fixes! I suppose that both me and @jonathandturner are kind of busy right now, but it would be great if you can contribute!

PR #87, opened 2019-09-18
Feat: Rust 2018 + rustfmt

timfish:
- Converts to Rust 2018 edition and fix all linting issues 
  - `use x::` -> `use crate::x::`
  - `try!(...)` -> `...?`
  - `Trait` -> `dyn Trait`
  - `'0'...'9'` -> `'0'..='9'`
- Latest rustfmt every file with default style
- Convert all example usages of `let mut scope: Scope = Vec::new()` -> `let mut scope = Scope::new()` so `Scope` backing type could be changed without breaking users code

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "rhai"
 version = "0.9.1"
+edition = "2018"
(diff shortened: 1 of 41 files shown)


jonathandturner:
Missed the email about this - sorry for the delay.

Looks good!

PR #89, opened 2019-09-30
Feat: Use std::any::type_name

### LINKED ISSUE #88: Use std:any::type_name rather than register_type_name
Rust 1.38.0 [now supports](https://doc.rust-lang.org/beta/std/any/fn.type_name.html) `std:any::type_name` which means `engine.register_type_name` could eventually be removed.

timfish:
Type names are now fully qualified. This could give bad ergonomics for arrays...

Fixes #88


jonathandturner:
Looks good. Thanks! 👍

PR #97, opened 2020-02-23
Handle arrays properly, plus other enhancements.

schungx:


schungx:
This PR does the following:

- Out-of-bound array access now comes back with an error instead of panicing
- `for` statement is added (shamelessly lifted from `stevedonovan/rhai`)
- `range` function is added to be used with `for` statement (shamelessly lifted from `stevedonovan/rhai`)
- `register_box_fn` added to `Engine` to register functions that return a `Box<dyn Any>` (using `register_fn` will box the `Box`.
- Augment arrays with `len`, `push`, `pop`, and `shift` functions.
- Expand the standard library functions with `print`, `debug`, and string concat operators.

jonathandturner:
@schungx this looks great! Thanks for all the effort you put into it.

PR #98, opened 2020-02-25
A whole bunch of enhancements.

schungx:
This PR does the following:

- Update `README` with new features
- Allow user to override the `print` and `debug` functions by passing closures to `on_print` and `on_debug`
- Introduce the `Dynamic` and `Array` type aliases to simplify API.
- A range of new built-in utility functions.
- Fixes a bug with unterminated strings.
- Make the `char` character type a first-class citizen with string indexing support.
- Boolean operators `&&` and `||` now short-circuit, while `&` and `|` don't.
- Else-if control flow.
- Comparison operators always return `false` if both operands are not of the same type, or if the comparison function cannot be found.
- `Engine` now cleans itself up after each `eval` or `consume`.
- Script-defined functions can override built-in functions.
- Comprehensive error reporting, line number / character position for evaluation errors.
- Bump version to `0.10.1`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rhai"
-version = "0.9.1"
+version = "0.10.0-alpha1"
 edition = "2018"
 authors = ["Jonathan Turner", "Lukáš Hozda"]
 description = "Embedded scripting for Rust"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,8 @@
 [package]
 name = "rhai"
-version = "0.10.0-alpha1"
+version = "0.10.1"
 edition = "2018"
-authors = ["Jonathan Turner", "Lukáš Hozda"]
+authors = ["Jonathan Turner", "Lukáš Hozda", "Stephen Chung"]
 description = "Embedded scripting for Rust"
 homepage = "https://github.com/jonathandturner/rhai"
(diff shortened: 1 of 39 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rhai"
-version = "0.9.1"
+version = "0.10.0-alpha1"
 edition = "2018"
 authors = ["Jonathan Turner", "Lukáš Hozda"]
```
jonathandturner: @schungx - you're doing enough, you should add yourself to the authors

schungx:
Thanks for making this crate.  It is exactly what I need for my own project, and so I'm eager to contribute back to it.

schungx:
I think this is about it, chock full of error reporting and diagnostics. Most features that are reasonably useful without making the language too complicated are added.

See if you can scan thru it one final round.  If you don't see any prob, I'll go ahead and press "Merge".

jonathandturner:
@schungx looks good to me

Do you want to try to publish the new version? I can add you as an owner on crates.io

schungx:
> 
> 
> @schungx looks good to me
> 
> Do you want to try to publish the new version? I can add you as an owner on crates.io

I guess better you do the honors on crates.io!

jonathandturner:
@schungx - it's now published

PR #103, opened 2020-03-04
Add no_std support

### LINKED ISSUE #19: How to use with no-std
I'm trying out rust on a cheap microcontroller following this tutorial https://polyfractal.com/post/rustl8710/

The tutorial works great and the sample echo program runs over serial.

I'd love to add in a rust scripting language and I really like the design of Rhai.  But when I add the library, the compile fails because of std is missing.

```
info: using existing install for 'nightly-x86_64-unknown-linux-gnu'
info: override toolchain for '/home/tim/rustl8710/src/rust' set to 'nightly-x86_64-unknown-linux-gnu'

  nightly-x86_64-unknown-linux-gnu unchanged - rustc 1.16.0-nightly (4ce7accaa 2017-01-17)

cd src/rust && xargo build --target thumbv7m-none-eabi
   Compiling rhai v0.4.0
error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv7m-none-eabi` target may not be installed
```

I'm fairly new to rust still and am not quite clear on what this all means.  Is Rhai meant to be used in such embedded use cases?  It sure would be neat if it or something like it did work.

VictorKoenders:
This adds no_std support to Rhai. The dependant crates are responsible for making sure the `alloc` crate is enabled. I've used https://github.com/rust-embedded/alloc-cortex-m for this.

The reason I chose for a feature called `no_std`, instead of adding a default feature `std`, is because we need to import several crates when we're *not* compiling from std. In most cases we have less dependencies.

- [HashBrown](https://docs.rs/hashbrown). HashMap is not available in alloc, because it uses an OS-based hasher by default: https://github.com/rust-lang/rust/issues/56192
- [core-error](https://docs.rs/core-error). Rhai uses `std::error::Error`, which is locked to std because of methods that are now deprecated. Core-error aims to be a replacement crate.
- [libm](https://docs.rs/libm). `core` does not have support for floating point mathematics by default, and there are 2 places where `f64::pow` and `f64::powi` are called. Libm provides those functions on different platforms.

Two new features are added in this PR:
- `no_std`: this enables `#![no_std]` and switches all imports to use `core` and `alloc`.
- `default_buildins`: This enables all the methods in `src/builtin.rs`. This feature is enabled by default. If users of this crate (including myself) are limited for space on their target platform, disabling this feature saves a nice 80KB (out of 155KB).

Closes #19

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -ex
+
+cargo build --verbose
+cargo test --verbose
+
+if [ "$TRAVIS_RUST_VERSION" = "nightly" ]
+then
+    cargo build --verbose --features no_stdlib
+    cargo test --verbose --features no_stdlib
+fi
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,5 @@ rust:
 matrix:
   allow_failures:
     - rust: nightly
+
+script: bash ./.ci/build.sh
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,14 +15,34 @@ include = [
 ]
 
 [dependencies]
-num-traits = "*"
+num-traits = "0.2.11"
+
+[dependencies.libm]
+version = "0.2.1"
+optional = true
+
+[dependencies.core-error]
+version = "0.0.1-rc4"
+features = ["alloc"]
+optional = true
+
+[dependencies.hashbrown]
+version = "0.7.1"
+default-features = false
+features = ["ahash", "nightly", "inline-more"]
+optional = true
+
+[dependencies.ahash]
+version = "0.3.2"
+default-features = false
+optional = true
 
 [features]
 #default = ["no_function", "no_index", "no_float", "only_i32", "no_stdlib", "unchecked", "no_optimize"]
 default = [ "optimize_full" ]
 debug_msgs = []     # print debug messages on function registrations and calls
 unchecked = []      # unchecked arithmetic
-no_stdlib = []      # no standard library of utility functions
+no_stdlib = ["num-traits/libm", "hashbrown", "core-error", "libm"]      # no standard library of utility functions
 no_index = []       # no arrays and indexing
 no_float = []       # no floating-point
 no_function = []    # no script-defined functions
--- a/src/any.rs
+++ b/src/any.rs
@@ -1,7 +1,8 @@
 //! Helper module which defines the `Any` trait to to allow dynamic value handling.
 
-use std::{
-    any::{type_name, TypeId},
+use crate::stdlib::{
+    any::{self, type_name, TypeId},
+    boxed::Box,
     fmt,
 };
 
@@ -12,7 +13,7 @@ pub type Variant = dyn Any;
 pub type Dynamic = Box<Variant>;
 
 /// A trait covering any type.
-pub trait Any: std::any::Any {
+pub trait Any: any::Any {
     /// Get the `TypeId` of this type.
     fn type_id(&self) -> TypeId;
 
@@ -27,7 +28,7 @@ pub trait Any: std::any::Any {
     fn _closed(&self) -> _Private;
 }
 
-impl<T: Clone + std::any::Any + ?Sized> Any for T {
+impl<T: Clone + any::Any + ?Sized> Any for T {
     fn type_id(&self) -> TypeId {
         TypeId::of::<T>()
     }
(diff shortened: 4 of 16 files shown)


VictorKoenders:
I got Rhai working on an stm32f103 with 64KB of FLASH. With this change the library was barely too big (77KB), however after temporarily removing some language features (who needs functions, variables and if-statements?) I managed to get Rhai down to 37KB.

https://twitter.com/victorkoenders/status/1235247234356310016

schungx:
> I got Rhai working on an stm32f103 with 64KB of FLASH. With this change the library was barely too big (77KB), however after temporarily removing some language features (who needs functions, variables and if-statements?) I managed to get Rhai down to 37KB.

@VictorKoenders   I have been thinking.  Why does cutting off functions, variables and if-statements cut down on code size?  AFAIK those features are quite simple to implement, and adds little code to the code base.  The significant elephant features are arrays, resolving array indexing, and custom types (with dot getters/setters chain).

I'd imagine that simply by cutting off custom types, you'd probably cut down 1/4 of the code size. Another probably 1/4 if you cut arrays.

However, maybe you need custom types to expose features in your embedded system... but those can easily be flattened into a bunch of function-based API instead.  So if it were for me, I'd be dropping custom types but keeping functions, because function definitions mostly use code that are common for evaluation function calls...

VictorKoenders:
Sorry my week's been busy. I'll finish the code before the end of the week.

> Why does cutting off functions, variables and if-statements cut down on code size? 

The large space saver is that the code ignores more tokens and has less complexity. The embedded version simply panics if it encounters a token it does not recognize. Because in my case all I had was a `loop` and 2 external fn calls, I didn't need to account for other language tokens.

From what I can tell, a large part of the rhai binary rhai is because of text-to-AST parsing. If that could be offloaded somewhere else, and have the AST send over and executed on the microcontroller, rhai would easily fit on a 64KB chip. In the future I can look into exporting and importing AST so this could be possible.

> I'd imagine that simply by cutting off custom types, you'd probably cut down 1/4 of the code size.

This is probably true. Disabling builtins already saved almost 50%. I think in my experiment I removed all variable assignment so rustc was probably able to optimize away most variables and types.

schungx:
OK, I did some experiments with the examples.  These program sizes are for standard release builds for x64 with standard library:

|Build|Size|Decrease|
|-----|-----|----------|
|standard|830K||
|unchecked|780K|50K|
|unchecked + no_index|680K|150K|

Therefore:

* checked arithmetic is 50K of code
* indexing feature is 100K of code

schungx:
OK, I did some experiments with the examples.  These program sizes are for standard release builds for x64 with standard library:

|Build|Size|Decrease|
|-----|-----|----------|
|standard|830K||
|unchecked|780K|50K|
|unchecked + no_index|680K|150K|
|unchecked + no_index + no_stdlib|560K|270K|

Therefore:

* checked arithmetic is 50K of code
* indexing feature is 100K of code
* the standard library is 120K of code

schungx:
So the major reduction ones are: arrays/indexing (as predicted), and floating-point support.

Your milage may vary because `no-std` doesn't have floating point in the first place.

schungx:
As expected, removing functions does not make the code size any smaller.

schungx:
I realize that by doing `lto=fat` and not setting `opt-level=z`, I have been optimizing for speed instead of size.

Rebuilding for size in 32-bits on a Windows system:

|Build|Size|Decrease|Feature|
|-----|-----|----------|------|
|standard|492K|||
|`unchecked`|464K|28K|`unchecked` = 28K|
|`unchecked` + `no_index`|408K|82K|`no_index` = 56K|
|`unchecked` + `no_index` + `no_float`|323K|169K|`no_float` = 87K|
|`unchecked` + `no_index` + `no_float` + `only_i32`|268K|224K|`only_i32` = 55K|
|`unchecked` + `no_index` + `no_float` + `only_i32` + `no_stdlib`|238K|254K|`no_stdlib` = 30K|
|`unchecked` + `no_index` + `no_float` + `only_i32` + `no_stdlib` + `no_function`|233K|259K|`no_function` = 5K|

VictorKoenders:
@schungx I've re-applied all the changes. I also moved all the imports to `stdlib.rs` which matches the `no_stdlib` feature you introduced.

With all the stuff going on in the world I'm not able to test this on a microcontroller any time soon. However seeing as this compiles and with the size optimalizations you've done, I'm certain it works fine. The codebase itself already worked and if people want to reduce the file size they can work it out.

PR #105, opened 2020-03-07
Fixes and cleanup plus an optimizer.

schungx:
- Improved documentation and code comments.  Minor code refactoring.

- Enhance `repl` and `rhai_runner` examples with error messages.

- Block expressions (a block returning the value of the last statement) can now be used wherever an expression can be used.  Only blocks right now, other statements cannot be used as expressions in order not to over-complicate the language.

- General code clean-up, avoiding unnecessary string allocations, and catch more errors at compile time (instead of evaluation time) with regards to LHS for assignments and indexing (which only takes non-negative integers).

- For dot-getter chain, allow arbitrary number of levels of indexing mixed in.

- Allow chaining arbitrary number of levels of dotting and indexing (but only one indexing at each level) for assignments with automatic propagation back.

- Avoid arithmetic panics by using checked operations - with the disadvantage of having to pull in the `num-traits` crate.  An added benefit is the ability to register _fallible_ functions (i.e. functions return `Result`) which gets mapped to `EvalAltResult::ErrorRuntime`.

- Arithmetic checking can be turned off by the `unchecked` feature.

- A rudimentary _optimizer_ to optimize the AST after compilation so subsequent evaluation will be faster.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,3 +20,4 @@ num-traits = "*"
 [features]
 debug_msgs = []
 no_stdlib = []
+unchecked = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,9 @@ include = [
     "Cargo.toml"
 ]
 
+[dependencies]
+num-traits = "*"
+
 [features]
 debug_msgs = []
 no_stdlib = []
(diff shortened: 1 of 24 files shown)


schungx:
The code handling dot-chains with indexing _really_ needs a cleanup...  it is way too complicated right now and unmaintainable.  When I get some time, maybe I'll start doing it...

PR #107, opened 2020-03-13
Make every file evaluation use PathBuf instead of str

Jarqwiz:
This is useful because filename might or might not be Unicode formatted.
This also helps the user if he/she doesn't want to use [to_str](https://doc.rust-lang.org/std/path/struct.Path.html#method.to_str) function when using Path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ use rhai::Engine;
 
 let mut engine = Engine::new();
 
-let ast = engine.compile_file("hello_world.rhai").unwrap();
+let ast = engine.compile_file("hello_world.rhai".into()).unwrap();
 ```
 
 Rhai also allows you to work _backwards_ from the other direction - i.e. calling a Rhai-scripted function from Rust.
--- a/src/api.rs
+++ b/src/api.rs
@@ -12,6 +12,7 @@ use std::{
     any::{type_name, TypeId},
     fs::File,
     io::prelude::*,
+    path::PathBuf,
     sync::Arc,
 };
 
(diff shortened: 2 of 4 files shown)


schungx:
One issue, the file API becomes a tad bit ugly to use:

~~~rust
 assert_eq!(engine.eval_file::<bool, _>("my_script.rhai")?, true);
~~~

The extra generic parameter is less nice than just the return type...

schungx:
How about taking a `Path` instead?

Jarqwiz:
Then you couldn't pass str straight to it.

schungx:
Well, that would mean the user has to create a `Path` object (or `PathBuf` or whatever)...  I suppose it is not difficult to create from a `&str`?  `Path::new(filename)`...

Jarqwiz:
It's true that API becomes more ugly, but I think it's still better to use generics instead of `Path`, because it allows users to use different parameters like `str` or `String`.

Jarqwiz:
Also it would be API breaking change if it takes `Path` instead of generic type.

For example
```rust
assert_eq!(engine.eval_file::<bool, _>("my_script.rhai")?, true);
```
will become
```rust
assert_eq!(engine.eval_file::<bool>(Path::new("my_script.rhai"))?, true);
```
or something

Jarqwiz:
OK!

I changed it to use `PathBuf` everywhere.
I think version number should be bumped, because this is API breaking change in some manner.

schungx:
> I think version number should be bumped, because this is API breaking change in some manner.

We haven't been publishing to `crates.io` as diligently as we should... so no worries!

Jarqwiz:
Ok. Thanks for info! :thumbsup:

schungx:
Looks good!  One final thing, maybe `filename` should be rename to `path` to indicate that it is a `PathBuf`.

Also line 170 in `README` will need to be fixed:

`let ast = engine.compile_file("hello_world.rhai")?;`

Jarqwiz:
Done!

schungx:
Looks good!

PR #110, opened 2020-03-19
eval

schungx:
This PR adds the _evil_ `eval` function for even more ways to shoot oneself in the foot.


schungx:
@jonathandturner   I believe this version is polished enough to warrant a new release to `crates.io`.

Thanks!

Issue #73, opened 2018-06-21
objects in rhai


brianwp3000:
I see documentation on how to define structs in Rust and then make them available for use in Rhai. These structs can have fields that are accessible with dot notation, e.g:
```
let foo = new_ts(); // new_ts() is a Rust-defined fn that returns a Rust-defined struct
foo.bar = 5;
```
Is there any way to create an object in Rhai without first having to define it in Rust? I'm looking for something like Javascript's objects which allow you do something like this:
```
var foo = { bar: 0 };
foo.bar = 5;
```

Is this on the roadmap for Rhai?

luciusmagn:
It's on the roadmap yes, but not possible yet, sorry

čt 21. 6. 2018 v 2:41 odesílatel Brian Porter <notifications@github.com>
napsal:

zutils:
I would also like objects in Rhai.  I REALLY would like to create structure types in Rhai and analyze them in Rust (viewing the properties/name/types etc...)

stevedonovan:
There is a fairly straightforward route here - first, provide support for maps as well as vectors. This seems to be easy enough, if the maps (HashMap) go from String to Any - then the indexing is unambiguous.  Second, allow 'object literals' like `{a = 1, c = "hello"}`.  The value of this is just `HashMap<String,Any>` which is easy enough to work with on the Rust side.

stevedonovan:
(Although that's syntactically ambiguous - I'm thinking too much in Lua here. But `{a: 1}` could work)

stevedonovan:
(Although that's syntactically ambiguous - I'm thinking too much in Lua here. But `{a: 1}` could work)

stevedonovan:
Sorry, comment lost: it is syntactically possible to have Javascript-style object literals like `{one: 1, two: "hello"}` which map to a `HashMap<String,Box<Any>>`.  Can unambiguously extend `[]` to do lookup on such maps.

jhwgh1968:
I've been playing around with Rhai for a project I'm doing, and found this as a roadblock. Just to try and understand the code, I started on a slightly different implementation:

```
let myarray = [0, 1, 2];
let mymap = [red: "value", orange: 4.0, yellow: 20];
```

Would this syntax be acceptable?

schungx:
> Would this syntax be acceptable?

I'm quite sure an object syntax is possible, even not hard to implement.  But is it really necessary to burden a _scripting_ language with objects?

[...]

PR #118, opened 2020-03-29
Basic HashMap/Object implementation

jhwgh1968:
This PR is my proposal, which is described in #73.

It adds basic support for an "object like" type, which to Rust is of type `HashMap<String, Dynamic>`. All keys are strings, and must follow the rules for identifiers currently enforced by the language.

The syntax is as follows:
```
let myarray = [
    1, 2, 3
];

let emptyarray = [];

let mymap = [:
    first: 1,
    second: 2.0,
    third: "three"
];

let emptymap = [:];
```

This type is hidden behind a new feature gate, `experimental_hashmap`, which is disabled by default. Unless explicitly opted in, the code behaves exactly as it does now. Unit tests for the new syntax are provided, and all previous unit tests pass.

The only thing you can do with these maps right now is create them, assign them, and return them. Indexing is not yet supported. The goal was to get feedback before I carried the implementation too much further.

Comments welcome.

cc @schungx @profan

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ no_stdlib = []      # no standard library of utility functions
 no_index = []       # no arrays and indexing
 no_float = []       # no floating-point
 no_function = []    # no script-defined functions
+experimental_hashmap = []    # new hash map type
 no_optimize = []    # no script optimizer
 optimize_full = []  # set optimization level to Full (default is Simple) - this is a feature used only to simplify testing
 only_i32 = []       # set INT=i32 (useful for 32-bit systems)
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -5,6 +5,9 @@ use crate::parser::{Expr, FnDef, Position, ReturnType, Stmt, INT};
 use crate::result::EvalAltResult;
 use crate::scope::{EntryRef as ScopeSource, EntryType as ScopeEntryType, Scope};
 
+#[cfg(feature = "experimental_hashmap")]
+use crate::parser::MAP;
+
 #[cfg(not(feature = "no_optimize"))]
 use crate::optimize::OptimizationLevel;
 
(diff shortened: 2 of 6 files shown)


schungx:
A few considerations:

* Choose a different operator pair for the literal?

Something like `#[` ... `]#` as I suggested in https://github.com/jonathandturner/rhai/issues/117

* Is it possible to directly cast a `MAP` to a custom type?  We know all the property getters and setters for each type, so technically it is possible.

```rust
engine.eval::<TestStruct>("let x = [: x:42 ]; x")?;
```

Technically speaking, we know that `TestStruct` has a property `x` which is registered as a setter with a certain type.  And the `MAP` contains a property called `x`.  If both types match, then we can downcast the `MAP` into `TestStruct` by first creating the struct (there is right now no way to register a default constructor) and then calling each setter.

schungx:
I would suggest adding a new `Token::HashStart` mapping to `[:` without whitespace between so your parsing function can be simplified even more.

In fact, if you do this, it doesn't really matter what `HashStart` is mapped to (we can change it in the future) and the parsing code does not need to be touched.

schungx:
I've put in a full implementation for object maps in my fork under the `object_maps` branch:

https://github.com/schungx/rhai/tree/object_maps

jhwgh1968:
Wow, a full implementation!

I looked it over, @schungx, and it looks good to me. In fact, I'd be willing to close this PR and use your branch instead once I make sure my prototype works with it.

I only chose the `[:]` syntax because it was the easiest way for me to hook into the parser. I think your `${}` syntax looks fine, and at least vaguely reminds me of Perl.

schungx:
Well, it wasn't too hard as all the mechanisms are already there.  Just need to add a few enum variants, match branches, then leech off the existing code for indexing and dotting.

It was a bit tricky to hijack property access for custom types for map-property access because the code was originally written exclusively with custom types in mind, so all property access gets compiled into getter function calls. In the end the solution was simple: hijack function call resolution instead, detect that we're really trying to get a property off a map, and Bob's our uncle.

I also like the feel of the `${}` syntax but I'm a bit concerned whether we'd be shooting ourselves in the foot when later, someday, we'll want to use the `$` notation (for strings interpolation maybe?).  I am struggling with myself whether `#{}` will be a better notation for a "hash" map...

schungx:
I'll close this PR if you have no objections.

jhwgh1968:
After poking at it for a while, I agree: your version is superior.

Closing in deference to #120.

PR #128, opened 2020-04-06
Reentrant Engine, Rust anonymous functions, `in` expression, timestamps, bug fixes

schungx:
This PR does the following:

* `call_fn` now has only one version and handles tuples of one/zero
* `optimize_ast` can now control the optimization level
* add `in` expressions
* add `timestamp` which is basically `std::time::Instant` to do rudimentary timing
* deprecates the `no_stdlib` feature - use `Engine::new_raw` to create raw `Engine`
* fix the max call-depth limit so it properly throws a stack overflow error instead of panic-ing
* fix bug in escaping single/double quotes in character/string literals
* `Engine` is now re-entrant
* support for creating anonymous functions in Rust which transparently calls into an `Engine` via `call_fn`
* Error enums are not omitted when certain features are turned on - this is to avoid the application needing to constantly uncomment/comment out match variants when using features.
* supports parsing JSON into an object map
* some more utility functions for arrays and object maps


jhwgh1968:
Small question: is there a better name than "anonymous functions"?

I always think of the lambda-like functions that rust has:

```
let x = |f| { f == 0 };
assert!(x(0));
```

I interpret this as more like "exported functions", where a named function in Rhai gets a function pointer in Rust.

(It might be useful for Rhai to support the syntax above, but that's a separate topic.)

schungx:
> I always think of the lambda-like functions that rust has:

Well, I thought about the name also.  However, `lambda` seems to imply that it somehow captures its environment as a closure, which this is not.  Rust's closures (basically lambda's) have a specific meaning.

This is more like a curryed function...

And this is what Rust's documentation says about anonymous functions:

https://doc.rust-lang.org/stable/book/ch13-01-closures.html

PR #133, opened 2020-04-11
Allow escaping quotes within a string

jhwgh1968:
Without this change, this is a syntax error:
```
let quote = "\"";
```

I'm surprised no one hit this before me.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -931,6 +931,11 @@ impl<'a> TokenIterator<'a> {
                     escape.clear();
                     result.push('\r');
                 }
+                // \"
+                '"' if !escape.is_empty() => {
+                    escape.clear();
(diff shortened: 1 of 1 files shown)


schungx:
Certain true!  Actually I'm about to merge a new PR https://github.com/jonathandturner/rhai/pull/128.  I'll port your changes to the new PR instead.

schungx:
Turns out that's not the place where the error is.  Here is the place where the escaped quote should be handled:

```rust
ch if enclosing_char == ch && !escape.is_empty() => result.push(ch),
```

It should be:

```rust
ch if enclosing_char == ch && !escape.is_empty() => {
    escape.clear();    // <- this is missing
    result.push(ch)
}
```

Issue #101, opened 2020-03-03
Benchmarking Rhai


schungx:
I think some formal benchmarks will be great.

I don't really have large/complex enough scripts to really do a meaningful benchmark. And I'm not really an expert in this... nor do I have access to proper system for this.

Furthermore, what is more important to measure?  Parsing speed?  Evaluation speed?  Memory/CPU footprint?

Suggestions?

KaiserKarel:
So if we'd be able to store ASTs; I'd consider parsing/optimization benchmarks separately from the actual engine benchmarks.

Some things I'd be interested in are:

1. Overhead of calling host functions (expensive computations can be offloaded to the host, if the overhead is low).
2. Some common benchmarks, such as Mandelbrot etc. These are easy enough to implement and give an indication of evaluation speed.
3. Memory and CPU are important as well; but unless we have a stdlib, or a set of prominent scripts to benchmark as part of the CI, we'd have little use for them? I'm guessing these benchmarks would be used to detect perf regressions.

With benchmarking, simply running them on your local laptop would provide little information, as the actual runtimes may wildly differ depending on the system it is run on. I'd say we create a small bench suite, and run those in the CI on PRs. That way we'd have some meaningful, reproducible feedback.

schungx:
1. Overhead of calling registered native functions should be very low, as essentially it is basically calling through a function pointer.  All of the overheads should be in setting up and copying the arguments.

2. I did put in a prime numbers sample, and it doesn't seem like Rhai is very fast compared to a compiled program.  However, it seems fast enough... not sure when running thousands of times though.

3. We can probably start with the prime numbers sample to detect regression...

KaiserKarel:
I can have a look at writing a PR this weekend, to add some benchmarks and possibly include them in CI.

schungx:
I figured that, since Rhai is treating every single value as a dynamic trait object `dyn Any`, requiring an allocation for _every single value_ that ever occurs in the script, it might be best to try to cut down on some of those allocations.  Otherwise large complex scripts are going to be heavily dragged down.

I have a new branch set up on my repo https://github.com/schungx/rhai/tree/enum-Dynamic that changes `Dynamic` to an internal enum that is essentially a union of most of the built-in types.  Other customer types are still represented by a `dyn Any` trait object.

[...]

PR #134, opened 2020-04-12
Initial benchmarks

KaiserKarel:
#101 
Created initial benchmarks using criterion, testing compile speed of a small script (primes.rhai), and the runtimes of primes.rhai and speedtest.rhai.

The primes benchmark is very slow tho; so we might consider setting the search threshold to 5000 instead of 10000.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ target/
 Cargo.lock
 .vscode/
 .cargo/
+.idea
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,3 +62,18 @@ optional = true
 version = "*"
 default-features = false
 optional = true
+
+[dev-dependencies]
+criterion = "0.3"
+
+[[bench]]
+name = "primes"
+harness = false
+
+[[bench]]
+name = "speedtest"
+harness = false
+
+[[bench]]
+name = "compile"
+harness = false
--- a/benches/compile.rs
+++ b/benches/compile.rs
@@ -1,0 +1,38 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+
+pub fn compile_benchmark(c: &mut Criterion) {
+    use rhai::Engine;
+    let mut engine = Engine::new();
+    c.bench_function("compile.rhai", |b|b.iter(|| engine.compile(black_box(PRIMES_RHAI))));
+}
+
+criterion_group!{compile, compile_benchmark}
+
+criterion_main!(compile);
+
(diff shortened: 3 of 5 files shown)


schungx:
Not sure about pulling in a new dependency just to do benchmarking...  Maybe `cargo bench` is enough?

Also, in addition to checking for regressions, maybe we need to also compare it with other scripting engines in other languages just so we know we're not _too_ slow?

https://github.com/jonathandturner/rhai/issues/101

GopherJ:
It'll be great if we can also add: https://github.com/rhysd/github-action-benchmark, so that we can visualize bench changes.

An example is: https://github.com/casbin/casbin-rs/pull/107

GopherJ:
Also it's better to use `test::Bencher` because seems criterion has some constraints especially on data format.

schungx:
> It'll be great if we could also add: https://github.com/rhysd/github-action-benchmark, so that we can visualize bench changes.

This thing looks really interesting... Not sure if it'd be difficult to set up...  I'm not too familiar with these GibHub Action thingies...

KaiserKarel:
In this PR, I chose not to add any benchmarking in the CI, as those are often unreliable. CI servers are quite noisy; which influences the benchmarks. If we add those; we should compare different CI providers to find the most reliable one.

As for the dependency; users should not notice, as it is only a dev-dependency. Criterion is the recommended standard, but I can alter the PR to use the regular cargo bench framework. The benchmarks are still run the same way, by using `cargo bench`.

KaiserKarel:
Combining criterion with crtcmp offers a really nice way to show benching differences. Criterion has different plotting backends as well.

schungx:
@KaiserKarel   if you can surf to https://github.com/schungx/rhai  see if you can start with the benchmarks I ported over under the `benches` directory.

GopherJ:
@KaiserKarel crtcmp cannot keep data.

schungx:
I have the benchmarks added in my own fork of the repo and now it seems to update successfully during each push.  Thanks @GopherJ !

I can't add it here because I don't have `Settings` access to this repo.  But judging from the results in my fork, it seems @KaiserKarel  is correct - micro-benchmarks on GitHub fluctuate quite wildly from run to run.  So, they can only detect really severe regressions because the measurements simply swing around way too much...

Running them on a development machine instead of GitHub appear to be stable.

schungx:
Some microbenchmarks using `cargo bench` are here: https://schungx.github.io/rhai/dev/bench/

Let's stick to no adding new crates for the time being...

PR #135, opened 2020-04-13
Implement string reverse method

jhwgh1968:
A simple implementation of `reverse` to handle some use cases in unpublished code, based on commentary.

Includes tests.

### Original Description
A fairly simple implementation akin to the version in the Rust stdlib, but with reverse slicing and without the ability to panic.

Includes tests.

### DIFF
--- a/src/builtin.rs
+++ b/src/builtin.rs
@@ -1025,6 +1025,9 @@ impl Engine<'_> {
             s.clear();
             s.push_str(&new_str);
         });
+        self.register_fn("reverse", |s: &mut String| {
+            s.chars().rev().collect::<String>()
+        });
         self.register_fn("trim", |s: &mut String| {
             let trimmed = s.trim();
 
--- a/tests/string.rs
+++ b/tests/string.rs
@@ -33,3 +33,23 @@ fn test_string() -> Result<(), EvalAltResult> {
 
     Ok(())
 }
+
+#[test]
+fn test_reverse() -> Result<(), EvalAltResult> {
+    let engine = Engine::new();
+
+    assert_eq!(
+        engine.eval::<String>(
+            r#"let x = "\u2764\u2764"; x.reverse()"#
+        )?,
+        "❤❤"
+    );
+    assert_eq!(
+        engine.eval::<String>(
+            r#"let x = "\u2764\u2764\u2764 hello!"; x.reverse()"#
+        )?,
+        "!olleh ❤❤❤"
+    );
+
+    Ok(())
+}


schungx:
I think this kind of feature would probably be best left as an external registered function?  If we go too overboard, we can implement the entire Rust `std` lib for all the primitive types...  I think we have to keep a balance on what minimal features will be frequently used by common scripts with bloating the Engine's standard library - remember they'll never get optimized away even if you don't use them, unless you exclude _all_ of them.

A normal user who wants `split_at` will most likely want to split on a delimiter character, unless he is dealing with fixed-width records.  He is likely to be more interested in a full-blown `split` function than `split_at`, which would require `index_of` or something similar to find the delimiter character in the string in the first place.

Actually, I would suggest that `index_of` and `sub_string` can be two good string functions to add, since you can easily build `split_at` from `sub_string` alone, and getting a sub-section of a string is going to be an interesting operation.

jhwgh1968:
I implemented this, because it would let me avoid implementing `split`, which turns out to be a little tricky and fairly expensive without streaming iterators.

Per your suggestion, I'll implement `sub_string` and `index_of`, and update this PR.

schungx:
I tried it myself, and it doesn't seem to be too straight-forward.  You have to be careful about the overflow errors (since Rhai uses `i64`/`i32` instead of `usize`).  Even if you discount that, it is not very useful to have `index_of` without a version that allows you to search from a particular character position... and that means translating between character positions and UTF-8 offsets.

schungx:
I have something working with `index_of` (two versions, one with a start character position), `sub_string` and `crop` (each with two versions, with or without length) in this commit:

https://github.com/schungx/rhai/commit/2e9a5f7a8945b2cce2b066c52caf467a3cb0eb29

jhwgh1968:
Two questions:

First, the version I was working on supported negative indexes from the back of the string:
```
let x = "abc";
let y = x.substring(x, -2, 2); // "bc"
let y = x.substring(x, -1, 1); // "c"
```

It looks like that doesn't work in your patch, because negative starting poinnts are forced to zero. Could that be added? There is code I'm looking at porting which makes heavy use of that.

Second, in addition to `index`, it would be good to have an `rindex` method -- or a `reverse` method of some kind in order to achieve that.

jhwgh1968:
Thinking about your comments of minimalism, I decided to implement `reverse`, which was easy. I can work around `rindex` and negative substring indexes with that.

Can this be merged, @schungx?

jhwgh1968:
Actually, I will close this and re-open it on your master instead, since it goes on top of your other changes.

schungx:
> negative indexes from the back of the string:

I think this is a great idea! This makes the whole API much more streamlined!

PR #136, opened 2020-04-13
add github action workflows

GopherJ:
Hello here some workflows which maybe useful. Please test. It should include:

1. Auto-release to crates.io and github releases when a new tag `v*` is created
2. Build and deploy docs to `gh-pages` branch
3. Add Github Actions Benchmark

Settings for **1** and **2**:

1. Add ssh deploy keys https://github.com/peaceiris/actions-gh-pages#1-add-ssh-deploy-key
2. Put your cargo token in [secrets](https://github.com/jonathandturner/rhai/settings/secrets) and name it as `CARGO_TOKEN`
3. Create `gh-pages` branch

If it goes well you should be able to see docs: http://jonathandturner.github.io/rhai/rhai

Settings for **3**:

1. Go to https://github.com/settings/tokens and create a token which `read` and `write` repo
2. Add this token in [secrets](https://github.com/jonathandturner/rhai/settings/secrets) and name it as `PERSONAL_GITHUB_TOKEN`

If it goes well and you have already some benchmarks you should be able to see it on: 
https://jonathandturner.github.io/rhai/dev/bench

Thanks for https://github.com/PsiACE who added these for me on casbin-rs. Hope it works for rhai too.


schungx:
1. I may be old-fashioned but I'd prefer to dry-run and double-check things before doing `cargo publish`...  things gone to `crates.io` stay there forever and cannot be taken back.

GopherJ:
@schungx  It's better to let CI do the check for you because it's repeat work. We just need to prepare correct unit tests.

GopherJ:
@schungx I think CI and code coverage is missing

schungx:
I suggest let's do one thing at a time... first put in the auto benchmarks to detect regression. 

But... I don't have settings access of `johnathandturner/rhai` so I can't put my token in there...

schungx:
@GopherJ  can I suggest that you rebase the PR on my fork so I can do the benchmarking there instead?

GopherJ:
@schungx  Like the above?

schungx:
@GopherJ  I'll need your help in this... not very familiar with how to work this GitHub Actions thingie.

I copied your benchmark.yml into my own repo and did the secrets thing as per your instructions.  Now how would I know the benchmarks are run and how would I see the reports?

schungx:
OK!  Success!  https://schungx.github.io/rhai/dev/bench/

GopherJ:
@schungx I'm also noob in github ~~~~ hhhh. I just copied what https://github.com/PsiACE  has done for me. Good to know it works!

udoprog:
To unstall this, any chance this could be separated into two PRs?

One that adds plain CI building (master branch, pull requests) with a matrix over features that should build. This is much easier to get done and shipped ASAP and would fully replace Travis. I'd be happy to take care of this. If everyone involved is OK with it, you can then rebase on top of it.

The other one that adds additional features: run benchmarks, publish docs, releases, etc...

schungx:
Absolutely doesn't mind this at all.  I know close to zero about everything you mentioned, so glad if you can pick it up.

udoprog:
Added building master branch and pull requests in #199

PR #150, opened 2020-05-06
Modules and speed improvements

schungx:
The big feature in this PR is modules support.


gmorenz:
First thoughts glancing through this code: 

Does this pull request make modules first class values? That strikes me as a poor idea for the future, it will make statically resolving functions substantially more difficult, which will substantially impede optimization.

I thought we had decided to allow only constants in modules, this looks like it has mutable variables? Or maybe that's an implementation detail?

I'm happy to see that we're adding user overridable index functions :)

gmorenz:
Continuing on on the module subject for a second - I think you might be thinking of these modules much like javascript objects. Unfortunately I don't think they can have the same performance characteristics, because of function overloading.

In javascript, a call to a function in a module/object is just an indirect function call, a single instruction that takes a value and sticks it into the instruction pointer. With runtime overloading we instead have to take the function name, look at the types, and decide what function to call.

Most of the time the types are probably known and we can do that statically, but not if we don't also know which function is being called. Modules as first class objects will make it much harder to know what function is being called.

schungx:
> First thoughts glancing through this code:

Yes, it does.

If you look into `Dynamic`, it has a variant that stores a `Module`.  It is _internal_ so users can't use it, but essentially it just wraps a `HashMap`.  Therefore, you can say modules are first-class.

> I thought we had decided to allow only constants in modules, this looks like it has mutable variables? Or maybe that's an implementation detail?

Right now all module variables are mapped as constants, so you can't change them from Rhai.

Unfortunately, you can always change a Rhai constant's _value_ through a Rust function.  You cannot change the constant variable, but you can change its value...

schungx:
> Continuing on on the module subject for a second - I think you might be thinking of these modules much like javascript objects. Unfortunately I don't think they can have the same performance characteristics, because of function overloading.

Yes, there will be a performance hit when searching a modules chain for the right function to call.

My thinking is that if performance is critical (or, in your case, you'd like to "static link" everything into bytecodes), then you disallow modules altogether.  The user can always concat a bunch of script files into a large script if performance is critical.  JavaScript has it this way also - in most production builds all modules are flattened into a single file.

schungx:
> I'm happy to see that we're adding user overridable index functions :)

Something I missed when using Rhai in my own project, and relatively easy to spin it in.  :-)

Unfortunately it is read-only at this point, as the code changes will be quite hairy for read/write...

gmorenz:
> My thinking is that if performance is critical (or, in your case, you'd like to "static link" everything into bytecodes) then you disallow modules altogether. The user can always concat a bunch of script files into a large script if performance is critical. JavaScript has it this way also - in most production builds all modules are flattened into a single file.

The primary purpose of modules isn't to have separate compilation units (indeed in rust they modules would all be in the same compilation unit), but to have namespacing.

I fully intend to support modules (namespaces) in some form for the users of my program, and currently it's looking like I'm going to have to write a separate likely incompatible implementation of them to get reasonable performance. This is disappointing because I don't think there is any reason that modules should have a significant negative performance impact (import statements going in and out of scope will come with a minor cost as things are renamed/hashes are changed. Other then that I would expect the performance impact to be 0).

Static/dynamic linking is a poor analogy for scripting languages. I fully intend to be building code that users type in at runtime and running it in the same millisecond. This is as dynamic as linking gets. I may even try to share the memory holding common libraries bytecode between many different "binaries" - though it will be a bit of a trick to implement that.

What I'm not particularly interested in is support for adding modules mid program run, that's just a more complicated way of saying "eval". And of course I understand that that behavior should be possible in general... by using "eval", it just shouldn't be the standard case.

The general attitude here is concerning to me too. Things don't easily become fast by doing an optimization pass afterwards, they become fast by being designed that way from the start. If performance isn't a goal of this programming language, that's fine and definitely you're decision, I'll go my separate way.  If so, might I suggest you change the documentation in the readme to reflect that? 

> Unfortunately it is read-only at this point, as the code changes will be quite hairy for read/write...

Should be really easy to do this with the get_mut proposal. Actually more work with those changes will be supporting get-only versions of indexing.

schungx:
> This is disappointing because I don't think there is any reason that modules should have a significant negative performance impact

That's true.  What we can do is, upon each `import` statement, immediately index all the module variables and functions, plus all of the sub-modules.  This way, calling a function incurs no more lookup cost than any other function.

I've thought about this, and it is possible to keep this index in the `State` type that is passed around.  I'll probably move also `fn_lib`, which contains all script functions, into the same type to be consistent.

> What I'm not particularly interested in is support for adding modules mid program run

It doesn't have to be conflicting goals.  We can be _both_ fast and support dynamic adding/releasing of modules.  It is just a bit tricky to implement.

>  they become fast by being designed that way from the start

True enough.  However, in most scripting cases, performance is not a critical bottleneck, and being flexible and easy is a quality by itself.

gmorenz:
Hey, so first off I think I might have come across as a bit too negative in the previous comment, sorry about that.

> It doesn't have to be conflicting goals. We can be both fast and support dynamic adding/releasing of modules. It is just a bit tricky to implement.

We absolutely can support that (which is the equivalent of... dlopen) without sacrificing performance in the linking before execution case - much like we do eval. Your description of how to handle import statements sounds correct, and it will come at minimal performance cost. Dlopen will probably never come at none when an optimizer exists, because you can't perform cross-linkage optimization (or at least it's very scary to think of trying to dynamically optimize the ast whenever the dlopen equivalent is called).

> True enough. However, in most scripting cases, performance is not a critical bottleneck, and being flexible and easy is a quality by itself.

Absolutely true, my use case is undoubtedly far more performance sensitive than most for scripting languages. To be perfectly honest the parts of the readme that talked about optimizing the size of values with feature flags are a large part of what got me interested in this language... that sort of optimization is pretty rare unless performance is a major goal. I am trying to make sure changes I push for in the language do not sacrifice substantial amounts of ease of use - but my bias is definitely towards performance.

schungx:
I am beginning to do some refactoring into making this happen, first by moving `fn_lib` into `State`, meaning that I now have a single point to search multiple places for function definitions.  That it would be relatively easy to support multiple sources.

schungx:
As a matter of fact, it turns out not to be difficult at all.  Just some code shuffling and refactoring.  Basically, when each module is loaded, I scan through all the modules and sub-modules and build hash all the possible module name chains.

For one module, it is almost as fast as a straight access because the variable lookup is by index into the current scope.  For more than one module (rare), the entire chain is hashed (pre-calculated) and then used to quickly seek the end module.  So, in other words, accessing a module-qualified variable or function can be _almost_ the same speed as accessing a variable in the scope.

schungx:
@gmorenz the latest commits https://github.com/schungx/rhai/commit/e6fabe58cc048b68eb2254debdaacfc147f39c0a change module lookup resolution to pre-computed hashes, making them essentially as fast as non-moduled access.

Thanks for your suggestions.  Now you don't have to worry about linking etc.

PR #163, opened 2020-06-13
Minor refinements and enhancements, WASM.

schungx:


alvinhochun:
It works on Wasm, but I need to edit the `Cargo.toml` and add the feature flag `wasm-bindgen` to the `instant` dependency. I think it would be best to follow the Readme of the `instant` crate to allow specifying the feature transitively by adding: 

```toml
[features]
stdweb = [ "instant/stdweb" ]
wasm-bindgen = [ "instant/wasm-bindgen" ]
```

schungx:
Too many features already, unless wasm is different enough to warrant its own feature, I'd rather not have another feature...

But I'll add `wasm-bindgen` to `instant`.

PR #171, opened 2020-06-25
Fix no_std build

stevefan1999-personal:


schungx:
You can simply add the imports into the `use crate::stdlib::{ ... }` import.  That maps automatically to the `no-std` alternatives.

For example:

`use crate::std::{string::ToString};`

schungx:
Thanks!

PR #176, opened 2020-07-05
Serde support, disable keywords, custom operators, custom syntax and chainable API.

schungx:


alvinhochun:
Why don't you merge individual features or changes back to the main repo as soon as they're done, but instead bundle them into one PR?

schungx:
> Why don't you merge individual features or changes back to the main repo as soon as they're done, but instead bundle them into one PR?

Well, basically to give it time for what I wrote to be tested out just in case there are bugs...

Anybody pulling from https://github.com/jonathandturner/rhai/ should expect a reasonable version which builds and runs correctly, not an experimental one...

PR #191, opened 2020-07-24
Avoid shadowing call errors in get_indexed_mut

udoprog:
I realize this is a compromise, but this shadowing the error made it really hard to figure out what underlying error caused the call to fail.

This simply allows whatever error that happens during the call to be propagated to provide better information.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1052,17 +1052,20 @@ impl Engine {
 
             #[cfg(not(feature = "no_index"))]
             _ => {
-                let type_name = self.map_type_name(val.type_name());
+                let val_type_name = val.type_name();
                 let args = &mut [val, &mut idx];
                 self.exec_fn_call(
                     state, lib, FN_IDX_GET, true, 0, args, is_ref, true, None, level,
                 )
                 .map(|(v, _)| v.into())
-                .map_err(|_| {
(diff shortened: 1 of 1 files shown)


schungx:
I know what you mean but this error is to show that an indexing operator cannot be used on types that do not support it.

Types that support it are:

* Strings
* Arrays
* Maps
* Any type with an indexer

A more proper way is to only shadow the error when trying to call an indexer comes back with `ErrorFunctionNotFound`.  This should allow all other errors through.

src/engine.rs
```diff
@@ -1058,12 +1058,6 @@ impl Engine {
                     state, lib, FN_IDX_GET, true, 0, args, is_ref, true, None, level,
                 )
                 .map(|(v, _)| v.into())
-                .map_err(|_| {
```
schungx, line 1061: A better solution is:

```rust
                .map_err(|err| match *err {
                    EvalAltResult::ErrorFunctionNotFound(_, _) => Box::new(
                        EvalAltResult::ErrorIndexingType(type_name.into(), Position::none()),
                    ),
                    _ => err,
                })
```

udoprog: Done, thanks!

PR #196, opened 2020-07-25
Avoid scanning over slices to validate numerical literals

udoprog:


src/token.rs
```diff
@@ -810,6 +810,32 @@ pub fn get_next_token(
...
+        'A'..='F' => true,
+        '0'..='9' => true,
+        _ => false,
+    }
+}
+
+/// Test if the given character is an octal character.
+fn is_octal_char(c: char) -> bool {
+    match c {
+        '0' | '1' => true,
```
schungx, line 826: Probably needs `'0'..'7'`?

udoprog: You are of course correct. I'll try and figure out why the test doesn't break before fixing it. Thanks!

schungx:
Can I merge this?  Have you checked the test?

src/token.rs
udoprog: Ok, so the reason it parsed is because when an escape sequence is encountered, it loops around and continues consuming numerical literals.

This actually makes me think that lexing of numericals might benefit from being simplified further: simply identify the kind of numeric literal and consumer everything alphanumerical until whitespace.

This appears to be how Rust does it: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c019d6741c5c4b6167bcc352be480ad1

For now I just fixed the immediate issue, thanks!

udoprog:
Yeah, feel free to merge. Just commented. Fixed it and hinted at future improvement to lexer: https://github.com/jonathandturner/rhai/pull/196#discussion_r460379025

src/token.rs
schungx: Not very comforting... because it should `break` out of the loop on the first invalid character...

But still, doesn't see anything wrong as yet, so merging this!

PR #198, opened 2020-07-25
Unbreak no std

udoprog:
This wasn't caught in CI since it's only tested on nightly, and nightly is [marked as an allowed_failure](https://github.com/jonathandturner/rhai/blob/master/.travis.yml#L7). But [it does indeed break](https://travis-ci.org/github/jonathandturner/rhai/builds/711679065).

You might also want to consider enabling running on pull requests!

Bonus change to bump `build.sh` from `sh` to `bash` syntax.

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -5,9 +5,7 @@ set -ex
 cargo build --verbose
 cargo test --verbose
 
-if [ "$TRAVIS_RUST_VERSION" = "nightly" ]
-then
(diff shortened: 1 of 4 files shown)


schungx:
Yes, I have a PR that fixes `no_std`.  Really, some automated CI will be great, as I have to constantly remind myself to run tests on `no_std` just to make sure nothing breaks.

PR #199, opened 2020-07-25
Build pull requests and master branch using Github Actions

udoprog:
Added some ~~basic~~ advanced feature matrixing.

Example run from my repo: https://github.com/udoprog/rhai/actions/runs/182120816 and build status below.


schungx:
Actually, I would suggest also include all the `no_XXX` features, each separately.  It is a constant pain that some code changes will break some `no_XXX` feature and not caught.

udoprog:
@schungx I hear you.

So that would result in a lot of build combinations, so would you be OK with only adding build feature combinations on `ubuntu-latest` and `stable`?

schungx:
> So that would result in a lot of build combinations, so would you be OK with only adding build feature combinations on `ubuntu-latest` and `stable`?

I would say that's fine because there really isn't any Linux-flavor specifics in the code base.  `stable` only is also OK, except for `no_std` of course...

.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
...
+    continue-on-error: ${{ matrix.experimental }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        flags:
+          - ""
+          - "--features no_std"
+          - "--features serde"
+          - "--features plugins"
+          - "--features unchecked"
```
schungx, line 22: OK, let's skip `unchecked`, `sync`, `no_optimize`, `only_i64` as they are less likely to have issues...

udoprog: `sync` is quite an involved feature, and it doesn't hurt to include them. It's more of an issue if we need to build a large collection of features for **x** platforms **x** toolchains.

But just adding three is not a big deal!

udoprog:
I've added `NoStdBuild`, which tries to build the [`no_std` project you are working on](https://github.com/jonathandturner/rhai/tree/master/no_std/no_std_test). I think that's a better test than trying to build the project. Note that these are marked as `continue-on-error: true`, since it doesn't seem to build on Linux or Windows right now.

I think this works for an initial integration if you feel like merging it. It already provides a lot of value and can be tweaked more later.

schungx:
> since it doesn't seem to build on Linux or Windows right now.

Not sure why not because it builds fine on my Windows box and Linux VM...  both release and debug.

.github/workflows/build.yml
schungx: Ah OK!

schungx:
@udoprog   I have encountered a strange thing.  The CI is working well so far...

I just published a new version, and it broke a lot of code.  I realize that certain feature flags have errors.

However, it is very strange because those feature flags are tested via CI and they all passed fine with no errors.

I wonder what is happening to cause this...

schungx:
Looking at the logs, all the tests seem to have the line: `running 0 tests`

So none of the tests are being run, for some reason...

udoprog:
Seems like there's a missing `$` in the build, meaning the variable is not being expanded:
https://github.com/jonathandturner/rhai/actions/runs/196429075/workflow#L53

I don't have the time to poke it right now, feel free to give it a shot!

PR #201, opened 2020-07-25
Fix building no_std on Windows and Linux

udoprog:
Trying to fix the `no_std` build.

[This suggests](https://github.com/rust-lang/rust/issues/47493) that the `lto` flag might be an issue. So I've made it configurable across platforms. At least this helps on my Linux box.

CC: #199 and @schungx

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,11 +54,13 @@ jobs:
   no_std_build:
     name: NoStdBuild
     runs-on: ${{matrix.os}}
-    # TODO: remove once build works.
-    continue-on-error: true
+    continue-on-error: ${{matrix.experimental}}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        include:
+          - {os: ubuntu-latest, flags: "--profile unix -Z unstable-options", experimental: false}
+          - {os: windows-latest, flags: "--profile windows -Z unstable-options", experimental: true}
+          - {os: macos-latest, flags: "--profile macos -Z unstable-options", experimental: false}
     steps:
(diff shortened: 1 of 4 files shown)


schungx:
Still failing on Windows...

udoprog:
Yeaaah, looking into it 😩

udoprog:
Can't easily figure out what's wrong with the Windows build. So I'm leaving it disabled for now. At least setting `lto = true` fixes the Linux build.

Feel free to keep poking at it.

schungx:
How about turning off `lto` on Windows?

udoprog:
@schungx So figured it out now. Removing lto gives more linker errors. Explanation follows.

`lto=true` in general eliminates the use of unused symbols, which allows it to build on Linux. This helps a bit on Windows, but apparently we either still require some crt items on msvc windows, or lto is unable to eliminate them.

The solution I've come up with is to manually link to the required crt libraries (math and crt).

I'm suspecting these are required regardless. Also note that I've attempted to use both [`+crt-static`](https://rust-lang.github.io/rfcs/1721-crt-static.html) and [`compiler-builtins`](https://github.com/rust-lang/compiler-builtins). The former seemingly doesn't do anything on Windows, the latter only provides some of the needed symbols (through the `"mem"` feature).

schungx:
Ah.  That's it.  I was using the GNU toolchain to compile on Windows.  Probably an MSVC thing.

Can you use GNU on Windows to get away from it?

udoprog:
Sure, but this change doesn't hurt imo since it allows it to compile and run on both. msvc is also default for Windows.

schungx:
Na.  Merge it!

PR #203, opened 2020-07-27
[WIP] Mutable closures for anonymous functions

Eliah-Lakhin:
**Note**: This is not ready for merging. I keep actively working on this feature, but would like to share a work-in-progress to hear some feedback, and to keep you in touch with the code changes since the final PR going to be quite large.

I will sync my branch with original `master` from time to time, so you don't need to care about my changes until they merged to upstream. But if the original `master` would be not too breakable that would save me time of rebasing/merging conflicts resolutions in my branch  :)

## Summary

Collecting all variable access attempts in anonymous function body during parsing for variables that are not explicitly declared(through the `let` keyword) considering that they are "closures" candidates and prepend them to the function's argument list(implicitly). Then implicitly `curry` these arguments in place in the parent context with the parent's context scope propagating this arguments the same way to the ancestors contexts recursively. In the end turn all `let` declarations affected by curried closures to become runtime-sharable Dynamic instances(through `Arc<RwLock<Dynamic>>`/`Rc<RefCell<Dynamic>>` depending on the `sync` feature).

The final goal is having a syntax like this:
```
let x = 100;

let func_1 = |z| {
  let y = 200;

  let func2 = || x + y + z;

  return func2.call();
};

|| {x = 50}.call();

func_1.call(30); // returns 280
```

## Implementation details

 - [x] Implement collecting of the closure candidates inside anon function body and curry them in place in the parent's context. This is done through extending of the `ParseState` object. The object is now holding a `externals: &'s mut Vec<String>` field that is enriching every time the parser detects variable access expression. I replaced `ParserState::find_var` with `ParserState::access_var` function that does the work of collecting under the hood. Once the parser finishes `Parser::parse_anon_fn` work it optionally enriches function's parameters and turns resulting expression(`FnCall`) to `curry` call expression depending on the current `ParserState::externals` list. And then propagates curried arguments to the parent context through the `ParserState::access_var`. This feature can be opted-out completely using a new `no_closures` feature.
 - [x] Introduce a new `Dynamic`'s `Union` variant called `Shared` that holds `Arc<RwLock<Dynamic>>` or Rc<RefCell<Dynamic>> (depending on the `sync` feature enabled). This variant can be created through the new `Dynamic::into_shared` function that ensures that the end-user doesn't wrap Shared into Shared multiple times. Since we need transparent interoperability between all `Dynamic` instances, but the `RwLock`/`RefCell` normally requiring read-write lock guards(`RwLockReadGuard`/`Ref` and `RwLockWriteGuard`/`RefMut`) instead of a normal `&` `&mut` references I had to change the `Dynamic` access interface by turning `downcast_ref` and `downcast_mut` to private and providing two new methods `Dynamic::read_lock` and `Dynamic::write_lock` that return `DynamicReadLock` and `DynamicWriteLock` lifetime-bound instances accordingly with `impl Deref`/`impl DeefMut` implementations for transparent access of underlying references. For normal `Dynamic` instances these objects just hold the same references as `downcast_ref` and `downcast_mut` would normally return and don't produce much overhead in contrast to prior API version. For Shared Dynamic they hold underlying appropriate `RwLock`/`RefCell` guards and return a result of `downcast_ref`/`downcast_mut` applying on underlying `Dynamic` once `deref()`/`deref_mut` is called. **This item is a breakable change**.
 - [x] Replace all `downcast_ref`/`downcast_mut` occurrences with the `read_lock`/`write_lock` functions accordingly including `fn_register` macros.
 - [x] Introduce a new `shared` built-in Rhai-script function that turns any expression's result into Shared Dynamic. This is done by introducing a new `KEYWORD_SHARED`. This feature can be opted out with the new feature `no_shared`, so the user will not be able to create Shared Dynamic explicitly.
 - [ ] Improve `Engine` such that it can interpret Shared Dynamic operands in left-hand-side properly by mutating underlying data instead of introducing copies to the current scope Scope. This is partially done, but I still have **difficulties** with call-chains and indexing-access syntax due to issues in introducing of `DynamicReadLock` and `DynamicWriteLock` to the `Target` variant since these instances are lifetime-bound to original Shared Dynamic. Perhaps, it will require a lot of work inside the `Engine`'s expression interpreter.
 - [ ] Use `ParserState::externals` information to implicitly turn all `let x = expr;` occurrences into `let x = shared(expr);` so that implicitly curry-ed variables will become mutable closures transparently for the script user. **Difficulties**: The information required for this turning is already gathering by the Parser, but I'm not quite sure how to change already constructed AST nodes. Perhaps, I can put a mutable bool flag(how? through the `Mutext`/`RefCell`?) to each `Stmt::Let` node to change it afterwards that the `Engine` will use to apply `Dynamic::into_shared`.
 - [ ] Provide appropriate tests and put according changes to the Rhai Book. Rust-doc comments already provided for the new API functions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ only_i64 = []       # set INT=i64 (default) and disable support for all other in
 no_index = []       # no arrays and indexing
 no_object = []      # no custom objects
 no_function = []    # no script-defined functions
+no_closures = []    # no automatic read/write binding of anonymous function's local variables to it's external context
 no_module = []      # no modules
 internals = []      # expose internal data structures
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ no_index = []       # no arrays and indexing
 no_object = []      # no custom objects
 no_function = []    # no script-defined functions
 no_closures = []    # no automatic read/write binding of anonymous function's local variables to it's external context
+no_shared = []      # no explicit shared variables in the script code
 no_module = []      # no modules
 internals = []      # expose internal data structures
 
--- a/src/any.rs
+++ b/src/any.rs
@@ -1,6 +1,6 @@
 //! Helper module which defines the `Any` trait to to allow dynamic value handling.
 
-use crate::fn_native::{FnPtr, SendSync};
+use crate::fn_native::{FnPtr, SendSync, SharedMut};
 use crate::parser::{ImmutableString, INT};
 use crate::r#unsafe::{unsafe_cast_box, unsafe_try_cast};
 
@@ -18,7 +18,14 @@ use crate::stdlib::{
     boxed::Box,
     fmt,
     string::String,
+    ops::{Deref, DerefMut}
 };
+
+#[cfg(not(feature = "sync"))]
+use crate::stdlib::{rc::Rc, cell::{RefCell, Ref, RefMut}};
+
+#[cfg(feature = "sync")]
+use crate::stdlib::sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard};
 
 #[cfg(not(feature = "no_object"))]
 use crate::stdlib::collections::HashMap;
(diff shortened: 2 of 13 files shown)


Eliah-Lakhin:
@schungx Would like to hear your opinion. Maybe it would be reasonable to split this work into two PRs? The base infrastructure neede for this feature is already provided in this PR. If I fix tests(some of them are currently broken for a reason) we can merge it first, and then I will implement the rest of the work in separate PR. What do you think?

schungx:
Thanks for the work!

Why don't I open up a new branch for you in my repo https://github.com/schungx/rhai and you can sync to that instead.

This way, I can make sure that I keep that branch in sync whenever I do some house cleaning on the `master` branch, and you won't get a godzillion merge conflicts.

schungx:
>  Use ParserState::externals information to implicitly turn all let x = expr; occurrences into let x = shared(expr); so that implicitly curry-ed variables will become mutable closures transparently for the script user. Difficulties: The information required for this turning is already gathering by the Parser, but I'm not quite sure how to change already constructed AST nodes. Perhaps, I can put a mutable bool flag(how? through the Mutext/RefCell?) to each Stmt::Let node to change it afterwards that the Engine will use to apply Dynamic::into_shared.

Well, the parser keeps a _state_, which contains a _stack_ of variable names.  It uses this to turn named variable access into a straight offset index.  For what you want to do, you might maintain a link between this variables stack and the corresponding `Stmt::Let` or `Stmt::Const`.  Then at the end of parsing, we run an extra pass over all the statements and fix up all the statements.  This can be done before the optimization pass, which does things in essentially the same manner.

schungx:
Alternatively, we just handle this in a different manner:

We introduce a new keyword, `shared`, that directly declares a variable as shared.  E.g.:

```rust
let shared x = 42;
let y = 123;

let f = |z| x += z;    // captures 'x'
let f = |z| y += z;    // 'y' is not captured
```

Then we can just add a `Arc<RwLock<T>>` / `Rc<RefCell<T>>` variant to `Dynamic` which will also solve the problem @jhwgh1968 is having in his plugins work.  See this PR https://github.com/schungx/rhai/pull/14

Eliah-Lakhin:
@schungx Thank you for review!

Honestly, in the beginning I thought it would be easier to implement shareable scope state then it appears to be when I start digging deeper inside the code.

The primary problem with wrapping `Arc<RwLock<?>>`/``Rc<RefCell<?>>` inside Dynamic is interoperability with other variants of `Dynamic`, and that issue lies down the Rust inconsistency between RwLock/RefCell access that is happening in runtime and the normal Rust static types borrowing that is checking in compile time. This issue raises a lot of questions when we are dealing with `downcast_ref`/`downcast_mut` which are strongly depend on the static lifetimes checking. We can't just leak the references of inner data laying inside RwLock or RefCell. They require to create explicit access guards creation(`Ref`/`RefMut` and `RwLockReadGuard`/`RwLockWriteGuard` accordingly) that mimic rust's borrowing rules in runtime. So there is actually no way to simply deference these containers inside downcast_ref/mut and return a reference to it, because the references that we are obtaining from their guard objects are depending to the lifetime of the guards themselves that we have to crate somewhere to do this dereference.

That was the primary reason to alter the public API of `Dynamic` object. Otherwise we will not be able to work with RwLock/RefCell data transparently in Rhai scripts. The following code wouldn't be possible without guards<->static-references interoperability:

```
let x = shared(100);
x = 200;
```
Because for Rhai the object created within `shared` would be different to the usual `Dynamic<Unit<INT>>`.

Eliah-Lakhin:
@schungx To sum up, I'm not sure if it's a good time for such API alteration, but sooner or later we will have to do something like this if we want to introduce shareable state into Rhai. Maybe there is a better approach, not I don't know :)

Anyway... For now I can propose a smaller scope of this Pull Request that will just automatically grab closures from the outer context without the way of mutating them in the anon function body. That is something I actually already done in the current PR, so I can create a separate PR with the smaller scope. And we will decide what to do with the Shared Dynamics later. What do you think?

schungx:
Well, I wouldn't suggest giving up so easily...

There may not be too many places in the code that requires deferencing a `&mut` to a `Dynamic`.  And if you look at the `Target` type, it is there for exactly that reason.

For example, `Target` can hold a reference to a _character_ inside a string.  They way it does that is to hold both the string itself, the character, and the index.  `Target` takes the place of a normal `&mut` reference to the object that is mutable.  For everything else, values are passed as copies so just cloning the `Rc`/`Arc` is just fine.

The attack plan is to:

1. Add `Dynamic(Union::Shared(x))` variant.
2. `Dynamic::downcast_ref` and `Dynamic::downcast_mut` will only succeed on the lock guard object types.
3. Or, we redirect all calls to `Dynamic::downcast_ref` and `Dynamic::downcast_mut` to `lock_read`/`lock_write` when it is a shared.
4. Make a variant of `Target` that internally holds a lock guard.

This way, you can splice in a reference-counted, locked object just like it is another data type.

schungx:
> Anyway... For now I can propose a smaller scope of this Pull Request that will just automatically grab closures from the outer context without the way of mutating them in the anon function body. That is something I actually already done in the current PR, so I can create a separate PR with the smaller scope. And we will decide what to do with the Shared Dynamics later. What do you think?

Yeah, I think that might be a good idea.

You want me to create a new branch `closures` for you in my repo?  I diligently make sure these branches are synced with `master` as much as possible.

Eliah-Lakhin:
> You want me to create a new branch closures for you in my repo?

Yes, please create a new branch, I will cherry-pick all stuff from my branch related to Parser changes only and prepare a few tests, so we can merge it separately. And then I will try to work on a Shared Dynamics in more proper way as you suggested.

Thanks!

schungx:
BTW, after some more thoughts, I realize that `Dynamic::downcast_ref` and `Dynamic::downcast_mut` are not used anywhere in the `Engine` - they are mostly used when interfacing with external functions.

Therefore, it may not be too bad to replace them with `lock_read` and `lock_write`.

The only think I would suggest change is to make the wrapped value a simple `dyn Any`, instead of having it be yet another `Dynamic`.  Yes, it would be an additional redirection for simple primary types, but with the overhead of reference counting and locks, I believe this is acceptable.  This should simplify the the type much more, and you don't get into recursive chains of locks within locks etc.

schungx:
OK there is now a `closures` branch in my repo.

Eliah-Lakhin:
Per discussion above I'm closing this PR and moving it's commits to separated PRs:

https://github.com/schungx/rhai/pull/19
https://github.com/schungx/rhai/pull/20

PR #225, opened 2020-08-25
Decimal (NOT YET FOR MERGE)

ObsceneGiraffe:
This is a draft implementation of integrating a `Decimal` type that depends on `['no_float']` being set.

The `['no_float']` dependency is required because the `Decimal` type hijacks the float parser as the Rhai language has no support for typed literals. Without a typed literal, the parser has no way to decide whether to store a decimal literal as a `float` or a `Decimal`.

I use the `rust_decimal` library as the Decimal implementation as it is pure Rust and still being maintained.


doc/src/language/values-and-types.md
```diff
@@ -32,6 +33,8 @@ If only 32-bit integers are needed, enabling the [`only_i32`] feature will remov
 This is useful on some 32-bit targets where using 64-bit integers incur a performance penalty.
 
 If no floating-point is needed or supported, use the [`no_float`] feature to remove it.
+If decimal support is necessary, use the [`decimal`] feature, but be aware that decimal and float support cannot cco-exist in the same build.
```
schungx, line 36: `cco-exist`

doc/src/start/builds/minimal.md
```diff
@@ -31,7 +31,7 @@ Opt-Out of Features
 ------------------
 
 Opt out of as many features as possible, if they are not needed, to reduce code size because, remember, by default
-all code is compiled into the final binary since what a script requires cannot be predicted.
+all code except [`decimal`] is compiled into the final binary since what a script requires cannot be predicted.
```
schungx, line 34: Well, actually there are a few opt-in features, such as `serde`.  So this statement is not technically correct.

doc/src/start/features.md
```diff
@@ -17,6 +17,7 @@ more control over what a script can (or cannot) do.
 | `sync`              | Restrict all values types to those that are `Send + Sync`. Under this feature, all Rhai types, including [`Engine`], [`Scope`] and [`AST`], are all `Send + Sync`.                                         |
 | `no_optimize`       | Disable [script optimization].                                                                                                                                                                             |
 | `no_float`          | Disable floating-point numbers and math.                                                                                                                                                                   |
+| `decimal`           | Enable decimal number support (through the 'rust_decimal' create). This will disable floating point support.
```
schungx, line 20: `create` -> `crate`

src/any.rs
```diff
@@ -578,6 +593,15 @@ impl Dynamic {
             Ok(s) => return (*s).into(),
             Err(val) => val,
         };
+
+        #[cfg(feature = "decimal")]
+        {
+            boxed = match unsafe_cast_box::<_, rust_decimal::Decimal>(boxed) {
```
schungx, line 599: I suppose `Decimal` is `Copy` so it can use `unsafe_try_cast` to save on one boxing operation?

src/any.rs
```diff
@@ -692,6 +716,14 @@ impl Dynamic {
             };
         }
 
+        #[cfg(feature = "decimal")]
+        if type_id == TypeId::of::<Decimal>() {
```
schungx, line 720: Again, consider `unsafe_try_cast` if `Decimal` is `Copy`

src/fn_call.rs
```diff
@@ -1203,6 +1206,27 @@ pub fn run_builtin_binary_op(
         }
     }
 
+    #[cfg(feature = "decimal")]
```
schungx, line 1209: I don't know if a `Decimal` can overflow or underflow.  You may want to actually copy the `INT` code block instead of the `FLOAT` code block if you need checked arithmetic...  At least the division should trap divide-by-zero...

src/packages/arithmetic.rs
```diff
@@ -356,6 +362,19 @@ def_package!(crate:ArithmeticPackage:"Basic arithmetic", lib, {
         reg_sign!(lib, "sign", f64, f64);
     }
 
+    #[cfg(feature = "decimal")]
```
schungx, line 365: In later versions I've changed the output of `sign` to be `INT` regardless of the type.  That's to avoid unnecessary boxed trait objects.  So therefore you can just return 1/-1/0 as `INT` and do away with `num_traits::One`.

schungx (commented):
This is great!  Some comments in this review.

Cargo.toml
```diff
@@ -19,6 +19,7 @@ categories = [ "no-std", "embedded", "wasm", "parser-implementations" ]
 [dependencies]
 num-traits = { version = "0.2.11", default-features = false }
 smallvec = { version = "1.4.1", default-features = false }
+rust_decimal = { version = "1.7", default-features = false}
```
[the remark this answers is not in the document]
ObsceneGiraffe: My understanding is that `optional` is for platform specific dependencies no? See the [Cargo docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies)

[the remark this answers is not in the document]
schungx: Not just for platform dependencies.  If a crate is only used by a feature, then it can be marked optional so it doesn't get pulled in during a normal build.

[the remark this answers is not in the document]
schungx: Also, there is a serde feature for rust_decimal. Rhai's serde feature should include rust_decimal/serde in its features set.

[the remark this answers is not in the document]
schungx: And currently it doesn't build the `no_std` sample.  However, it could be due to some simple code duplication...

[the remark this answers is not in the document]
ObsceneGiraffe: Ah I see how to use optional dependencies now. :wink:

src/any.rs
schungx: Maybe just copy the `FLOAT` block:

```rust
        #[cfg(not(feature = "no_float"))]
        if type_id == TypeId::of::<Decimal>() {
            return <dyn Any>::downcast_ref::<Decimal>(&value)
                .unwrap()
                .clone()
                .into();
        }
```

schungx: Also, if you have `serde` enabled, it may attempt to serialize `f32` or `f64` into a `Dynamic` via `Dynamic::from`.  Therefore,  you'd want to also handle the case where 1) `no_float`, 2) `decimal`, 3) `serde` (which I suppose must have floats...), 4) the type is `f32` or `f64`.  In such a case, you should convert the floating-point number into `Decimal`.:

ObsceneGiraffe: Using `unsafe_try_cast` is possible, but this breaks the transfer of ownership pattern at play, where the `Result::Err<Box<...>` returned from `unsafe_cast_box` is used to move the boxed value back from the `unsafe_try_cast`.

ObsceneGiraffe: The problem being that `unsafe_try_cast` returns an `Option` meaning it consumes the boxed value. I'm unsure how to resolve that. Maybe by borrowing the contents of the Box to `unsafe_try_cast`?

schungx: Sorry, my mistake. You can't use `unsafe_try_cast` here.  Just `downcast_ref` and then `clone` as my example.

src/any.rs
schungx: `unsafe_try_cast` here should be fine!

src/any.rs
schungx: They key points are: 1) `Decimal` is `Copy` so cloning is cheaper than boxing (supposedly), 2) You should handle the rate conversion cases to/from f32/f64.

ObsceneGiraffe:
All good 😉 this will be ready tomorrow as my fence just blew over in a storm, so that’s first priority 😆

Brian

src/fn_call.rs
ObsceneGiraffe: I'll have to take a look at how this all fits together a bit closer. `Decimal` can overflow, but it has checked versions for add, sub, etc that return an Option. 

Rhai also has Checked traits that return an Option, so all good there. I'll see later on today if I can use those traits for the `Decimal` type.

src/packages/arithmetic.rs
ObsceneGiraffe: The num_traits::One usage has been removed, and replaced with straight 1, -1 literals.

src/any.rs
ObsceneGiraffe: `unsafe_try_cast` version is pushed.

src/packages/arithmetic.rs
```diff
@@ -365,13 +362,7 @@ def_package!(crate:ArithmeticPackage:"Basic arithmetic", lib, {
...
     {
         lib.set_fn_1("sign", |x: Decimal| {
-            Ok(
-                if x.is_sign_positive() {
-                    Decimal::one();
-                } else {
-                    Decimal::one().set_sign_negative(true);
-                }
-            )
+            Ok(if x.is_sign_positive() { 1 } else { -1})
```
schungx, line 365: If x == 0 then it should return 0, not 1...

src/any.rs
ObsceneGiraffe: `downcast_ref` version is pushed.

I have to look at the `serde` case closer later on. I need to understand the flow there. `rust_decimal` has `serde` support.

src/packages/arithmetic.rs
ObsceneGiraffe: Indeed, I'm being sloppy!

src/packages/arithmetic.rs
schungx: I've put in a comment: it should return 0 when x== 0.

src/any.rs
schungx: Yes.  I think it probably defaults to serialize to/from JSON as f64 or f32.  Thus you need to consider some of the plumbing in the `serde` feature that serializes to/from `Dynamic`.

However, it is not a big deal, as I can clean them up later on.

ObsceneGiraffe:
@schungx Ok I'm gonna be busy with adulting for the rest of the day. I'll come back to the open issues tomorrow morning. have a good one :wink:

src/packages/arithmetic.rs
```diff
@@ -362,7 +362,13 @@ def_package!(crate:ArithmeticPackage:"Basic arithmetic", lib, {
     #[cfg(feature = "decimal")]
     {
         lib.set_fn_1("sign", |x: Decimal| {
-            Ok(if x.is_sign_positive() { 1 } else { -1})
+            Ok(if x == Decimal::zero() {
+                0
```
schungx, line 366: You have to be careful about types here.    This may not be `INT`.

You'll need to do `0 as INT` etc.

src/fn_call.rs
schungx: Yes, look into the `add`, `sub` etc. functions.  Those are checked versions.  You can copy that template for your own `add_decimal` etc.  Essentially if it overflows, convert it into an error.

schungx: Well, actually I see that `Decimal` seems to implement `num_traits::CheckedAdd` etc.  If so, maybe you can simply use the generic versions of `add`, `sub` etc. and no need to write your own versions.

ObsceneGiraffe:
Hey hey, after an unexpected break I'm back in business ;)

schungx:
Great to have you back!  But a bit of a "doh!"...

I just merged in a major revision that implements a plugins system with procedural macros.

The files under `packages` have been changed significantly.  You may have to do some revision to merge into that.

Or you can leave it to me...  I can apply your changes to the new code base when you're ready.

Sorry!

ObsceneGiraffe:
Nah don't worry about it, I'll merge in the changes ;)

schungx:
@ObsceneGiraffe   let me know when this can be merged.

ObsceneGiraffe:
@schungx well all the tests are passing. I merged the changes in upstream yesterday. But the Serde work remains to be done. I'm starting that now.

ObsceneGiraffe:
I took a look at splitting the `gen_arithmetic_functions` marco up into gen_bitwise_functions  so that I could use that for defining the Decimal functions too. I had some trouble making the `if y == 0` test generic in the `fn divide()`.

Decimal implemented trait Zero but all the integer types do not for some reason?

But anyways, I'm going stash that work and get on with the serde / Decimal integration.

schungx:
> But anyways, I'm going stash that work and get on with the serde / Decimal integration.

No prob.

> I took a look at splitting the gen_arithmetic_functions marco up into gen_bitwise_functions so that I could use that for defining the Decimal functions too.

You can, but `gen_arithmetic_functions` currently is registered in three groups.  If you split it, then it doubles to six groups.

May not be worth it, as having a dedicated module like what you have is perhaps simpler.

> I had some trouble making the if y == 0 test generic in the fn divide().

This is why `Zero` is gone: https://github.com/rust-lang/rfcs/pull/369

I'd hesitated to pull in `num_traits` just to have it though...

ObsceneGiraffe:
> You can, but `gen_arithmetic_functions` currently is registered in three groups. If you split it, then it doubles to six groups.
> 
> May not be worth it, as having a dedicated module like what you have is perhaps simpler.

Yea the increase in groups is not worth the trivial duplication in the `decimal_functions` module. 



> This is why `Zero` is gone: [rust-lang/rfcs#369](https://github.com/rust-lang/rfcs/pull/369)

And good to know

ObsceneGiraffe:
On the serde support. I don't think the Dynamic type should be modified to support converting floats to Decimal, as that's an inherently dangerous operation in terms of loss of precision.

`rust_decimal` itself has implementations of `serde::Serialize` and `serde::Deserialize` traits. The `rust_decimal` implementations convert directly to and from string by default unless `serde_float` is enabled.

From what I see, I would have to integrate the `Decimal` impls for `serde::Serialize` and `serde::Deserialize` into the Rhai `DynamicDeserializer` and `DynamicSerializer` in `de.rs` and `ser.rs` respectively.

schungx:
However, I think the serde serialization trait requires you to implement a serializer for `f32` and `f64`...  I'm not particular fluent with serde, so maybe you can just `unimplemented!()` them.

ObsceneGiraffe:
Yep `serde` is new for me too. But that's where I'm heading with the `serde` support.

schungx:
A reminder: Rhai serde support is essentially a new serialization _format_.  `Dynamic` is a format, which things can serialize into.

That's why it is `ser::to_dynamic` and `de::from_dynamic` and not the other way round.

ObsceneGiraffe:
@schungx I'm having some trouble figuring out how to integrate the `serde` definitions in `rust_decimal` into the `Deserializer` impl in `de.rs:160`.

`rust_decimal` has various `serde::Deserialize` impls in `serde_types.rs` and it looks like I need to integrate that `Deserialize` impl into the `match` statement in `de.rs:160`. It's unclear to me what I need to do here. Should I somehow create a new instance of `rust_decimal::serde_types::DecimalVisitor`? Or should I directly call `Deserialize::deserialize` directly on the `Decimal` type.

My `serde` knowledge is quite limited, I could so with help from someone with more knowledge of `serde`. Can you put me in contact with someone to discuss this integration?

schungx:
I'm not an expert in `serde` also...  Actually I did the ground work and somebody else refined it...

Deserialization means going from `Dynamic` to `Decimal`.  Maybe it doesn't need any special treatment...  because `serde` only understands `f32` and `f64`...  So it can be `Dynamic` -> `f64` -> `Decimal`, but then you may have rounding errors...

ObsceneGiraffe:
Yea that's what I'm trying to avoid. I want to warn if someone is trying to deserialize from floats to Decimal. I explicitly want to only handle `string -> Decimal`

schungx:
If there is built-in support for `String` -> `Decimal`, then it is simple...

Just do:

```rust
Union::Decimal(_) => self.deserialize_str(visitor),
```

That would convert a `Dynamic` value with `Decimal` type into a string for `serde`.  Then, `serde` will convert it into your `Decimal` from that string, if that's what you plan to convert to...

schungx:
Decimal is now built into Rhai.  Retiring this PR.

ObsceneGiraffe:
Thanks @schungx, apologies, I disappeared into the wilderness for a while.

PR #228, opened 2020-08-29
Make stable versions better use SemVer

ocornoc:
Currently I'm having an issue using Serde with Rhai, and I found that it's because Rhai isn't taking advantage of Cargo's allowance of `^` for versioning.
I've added the caret for stable dependencies only.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ categories = [ "no-std", "embedded", "wasm", "parser-implementations" ]
 
 [dependencies]
 num-traits = { version = "0.2.11", default-features = false }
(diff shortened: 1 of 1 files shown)


schungx:
A semver without any operator is taken to be with a caret: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

schungx:
> Currently I'm having an issue using Serde with Rhai,

What is the problem you face with serde?

ocornoc:
I'll move this to an issue then, apologies.

ocornoc:
Actually, nevermind, I think I solved the issue.

PR #229, opened 2020-08-31
Plugins.

schungx:


timfish:
The plugins link in the readme is broken:
https://schungx.github.io/rhai/plugins/index.html

schungx:
Yes, good catch.  It is in https://schungx.github.io/rhai/dev/plugins/index.html right now.

PR #271, opened 2020-10-29
Fix lifetime issue in engine with `internals`

BrettMayson:
I was getting this error, and this fixed it for me
```
error[E0312]: lifetime of reference outlives lifetime of borrowed content...
   --> /.cargo/registry/src/github.com-1ecc6299db9ec823/rhai-0.19.3/src/engine.rs:466:9
    |
466 |         self.mods
    |         ^^^^^^^^^
    |
note: ...the reference is valid for the lifetime `'a` as defined on the impl at 454:19...
   --> /.cargo/registry/src/github.com-1ecc6299db9ec823/rhai-0.19.3/src/engine.rs:454:19
    |
454 | impl<'e, 'x, 'px, 'a, 's, 'm, 'pm, 't, 'pt> EvalContext<'e, 'x, 'px, 'a, 's, 'm, 'pm, 't, 'pt> {
    |                   ^^
note: ...but the borrowed content is only valid for the anonymous lifetime #1 defined on the method body at 465:5
   --> /.cargo/registry/src/github.com-1ecc6299db9ec823/rhai-0.19.3/src/engine.rs:465:5
    |
465 |     pub fn imports(&self) -> &'a Imports {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -455,7 +455,7 @@ impl<'e, 'x, 'px, 'a, 's, 'm, 'pm, 't, 'pt> EvalContext<'e, 'x, 'px, 'a, 's, 'm,
     #[cfg(feature = "internals")]
     #[cfg(not(feature = "no_module"))]
     #[inline(always)]
-    pub fn imports(&self) -> &'a Imports {
+    pub fn imports(&'a self) -> &'a Imports {
         self.mods
     }
     /// Get an iterator over the namespaces containing definition of all script-defined functions.


schungx:
Yes, good catch, but I'm about to land a new PR including this fix.  I'll close this, and the PR will land shortly once it goes through CI.

schungx:
FYI, the fix has landed.

Just out of curiosity, what do you need `internals` for?  If it is for some regular use, maybe I should expose them by default...

BrettMayson:
I want to validate the scripts before they are ran. I am using rhai to coordinate micro-services written in Rust. I want to check the AST and make sure that all the custom functions in the script actually exist on the micro-services before execution starts to prevent errors and unexpected states if flows only half complete before crashing. I'm currently trying to do something like
```
let numbers = service("number-service");
numbers.sum.call(1, 2);
```
I want to fetch the metadata from `number-service` and ensure it has a function called `sum` that takes two integers before any execution starts.

schungx:
Ah OK... that's probably a valid use of `internals`, but beware that the `AST` data structures have undergone a number of changes - I try to pack as many of them together as possible to improve cache-friendliness.

So your code may need changes, but luckily `Expr::FnCall` has not changed.  Most changes are in `Stmt`.

schungx:
@BrettMayson 

Ah, me and my big mouth.  Just after I said `Expr::FnCall` has not changed, I've changed it by extracting all the info into a new `FnCallInfo` struct - to prepare for possible future extension when function pointers can simply be called instead of having to use the `call` method.

It should be a relatively simple change.

PR #285, opened 2020-11-08
Extract variables

JPMoresmau:
This is to start to address https://github.com/jonathandturner/rhai/issues/253. I'm investigating rhai and I would need this functionality (to know which variables a script would require in the provided scope or variable resolver). Let me know if this is OK


schungx:
This looks OK (except for failing the fmt test, which should be simple to fix).  However, I hesitate to include this into the main repo because it looks to be quite specialized...

Can this code work outside of the Rhai repo itself, with `internals` turned on?

JPMoresmau:
Are Expr and Stmt enum values exposed even with internals on?

JPMoresmau:
OK in fact this code works fine in my project if I use internals, apart from the deprecation warning on AST.statements(). So ok if you don't want this code it's fine, I can live with that!

schungx:
It is better left out of the main repo, but we can put together a new crate, something like a `rhai_helpers` that contains these useful tools...

schungx:
Also, there seems to be a problem with your logic.  I think you're going to need a stack of variable definitions, and avoid adding to `vars` when a variable exists on the `defs` stack.

Under your current code, you're going to have problem with this:

```rust
let x = 1;

{
    let y = 42;
    print(y);     // here 'y' is local
}

print(y);    // here 'y' is global
```

Your code will see 'y' being in `defs` and remove it from `vars`.

You need to convert `defs` into a `Vec`, store `defs.len()` when entering a block scope, and then `defs.truncate(orig_len)` when you exit that block scope.

JPMoresmau:
Yes, this case is not handled, because I just needed a quick extraction of variables from expressions, not really statements, for my use case. Is that a common case? Anyway close this PR since it doesn't belong here, I'll work on the code for my use case and if there's a helper project I'll add the proper handling here

schungx:
> Yes, this case is not handled, because I just needed a quick extraction of variables from expressions, not really statements, for my use case

In that case, if you only use `compile_expression`, then you don't even have to handle all the `Stmt` cases.  It will only be `Stmt::Expr`.  There cannot be other statements.

bogzbonny:
Hey, @schungx is there any interest in actually having a `rhai_helpers` library? 
I actually need this ability for my usage of rhai. My usage includes a preprocessing stage where I need to set the scope of for an evaluation based on what's actually in the rhai script... in other words I need to access all the variable names so I can set the scope, hence I need to actually extract them from the AST. Additionally, I actually need this functionality for full scripts not just expression-scripts.

This code was useful to me (thanks @JPMoresmau) , I recently updated it for the rhai version 1.12.0

I put it up here just incase anyone else is in my shoes: https://gist.github.com/bogzbonny/8efacf3334172aba84e4d24c00ea45a8

This being said, I'm not super comfortable with all the inner workings of rhai, I was basically just going off of the types in an attempt to corral them into the HashSet, I wouldn't be surprised if there were actually some bugs in this code, hence I wouldn't mind having that code live in a designated `rhai_helpers` repo with a few more expert eyes on the matter :)

schungx:
Yes, that would be useful, although some care must be taken to handle local shadowed variables.

I can see a function that scans an AST and returns all the "external" variable accesses so that a user can prepare them in a `Scope`.

I would probably base it upon the `walk` API of the AST, which mainly does what you're doing.  You can provide a callback `AST::walk` and then match on `Expr::Variable`...

You'd also need to trap all the `let`/`const` definitions in a stack so you can avoid local variables, for example:

```rust
{    // start block
    let x = 42;    // stack x
    print(x);        // x is not an external access
}    // unstack x at end of block
print(x);    // now x is an external access
```

schungx:
You can take some inspiration from `Engine::compile_into_self_contained` which uses a similar tactic to "capture" all `import` statements and "pre-resolve" them.

bogzbonny:
Cool, yeah I understand and like your proposed design. Just to confirm, you can determine each local scope as all scope-bound statements are grouped together within the `StaticVec` of a `StmtBlock` object.. Correct? 

I'll open a PR for this when I start working on this feature... might not be for a month or so as I've got a heavy workload, but it's in my work queue.

schungx:
Well, you get a stack of the AST nodes up to the current node being visited... so technically you can inspect it to determine the scope bounds by matching on the `StmtBlock` node.  However, this seems a bit convoluted to me...  and I am not 100% certain (I am 90%+ certain though) that `StmtBlock` will catch all of them...

I can suggest a simpler method:

* start with an empty array which contains tuples of `(String, &Stmt)`, the `String` should be the name of a variable/constant defined using `let` or `const`, and the `Stmt` is the one containing the definition statement.

* when encountering `let` or `const`, push the variables to the array with the current container

* when visiting a node, always find its container in the array.  Anything _after_ the container in the array should be removed (they have gone out of scope.

Originally I was hoping to use levels counting, but then I realize that statements such as `if` will jump from the then-clause to the else-clause directly, maintaining the same path depth but changing scopes.

bogzbonny:
interesting point about the if-else jumps. 
I think I get your approach besides that I don't yet understand how I would "find its container in the array". 
Are you referring to find the container `Stmt` within the tuple array? Does this mean that the tuple array contains all `Stmt`s ; meaning that elements which are _not_ a `let` or `const` statement are still added to the tuple array as a reference for when searching for the container of a future node?

schungx:
> I don't yet understand how I would "find its container in the array".

What I mean is to search in `path` backwards until you find a `StmtBlock`... I suppose that's the only place that scopes are created...  The closest `StmtBlock` to the current node is its scope container.

If you don't find any, then the statement is at top level.

PR #320, opened 2020-12-30
engine(module_resolver): Fix build for no_std

KokaKiwi:
Initialize a dummy module resolver instead of `None`, as in `Engine::new_raw()`

Related change: https://github.com/jonathandturner/rhai/commit/dc4e52e7954db4f52da0cbb917806961003c9eb6#diff-ff278ec599adfd033d35a083ebe8948db8bf177c9817dee8ddc4f3a9e3f5728eR767


schungx:
Ouch!  I better change my CI example of `no_std` to catch this.

PR #326, opened 2021-01-07
Fixed an issue when compiling no_std and no_optimize on a target with no alloc::sync

VictorKoenders:
I'm trying to run rhai on a microcontroller called [atsamd21g18](https://docs.rs/atsamd21g/0.9.0/atsamd21g/). This is a thumbv6m microcontroller. This target does not have the `alloc::sync` namespace: https://doc.rust-lang.org/stable/src/alloc/lib.rs.html#179-180

image

I import rhai without the `sync` feature, and the `stdlib::sync` import doesn't seem to be used, so I put it behind the `sync` feature flag.

### DIFF
--- a/src/stdlib.rs
+++ b/src/stdlib.rs
@@ -11,7 +11,9 @@ mod inner {
     #[cfg(not(target_arch = "wasm32"))]
     pub use core::{i128, u128};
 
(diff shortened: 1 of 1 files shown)


schungx:
Thanks for catching this.

PR #349, opened 2021-02-11
[WIP] [feature request] Add scientifc notation literal

Kezii:
I often use scientific notation float literals (123e4 == 123×10**4), but it's missing in rhai

This implementation is wrong (I added it as an operator and not a float literal ... ?) but the concept is there

If you want, delete all of this and do it properly


schungx:
Yes it is probably better to capture it as part of a float literal. Closing this for now. I'll try to put it in.

schungx:
Scientific notation is pushed to master: https://github.com/rhaiscript/rhai/pull/350

PR #389, opened 2021-04-06
add test for f32 + serde

dmadisetti:
This fails locally

image

seeing if it breaks the action too

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,7 @@ jobs:
           - "--features no_optimize"
           - "--features no_float"
           - "--features f32_float"
+          - "--features f32_float,serde"
           - "--features decimal"
           - "--features no_float,decimal"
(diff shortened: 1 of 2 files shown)


dmadisetti:
`et voila`

@schungx just an FYI. This might be a trivial fix- I haven't looked at it at all. I might find time either tonight or tomorrow night (NY time)

dmadisetti:
Woops. That might just be the tests though. Anyway, leaving this for later

schungx:
Yes, thanks for catching this.  The combination of `serde+f32_float` pops out this typo.

With the number of feature flags, it sometimes get hard not to miss certain combinations...

I'll fix the typo directly in the code and add `serde,metadata,internals` to `f32_float` in CI.

schungx:
Fixed by https://github.com/rhaiscript/rhai/pull/390

PR #514, opened 2022-01-26
Fix Unix CI for no-std

Friz64:
Took a bit of digging until i found the issue i was looking for.

https://github.com/rust-lang/rust/issues/79609

### DIFF
--- a/no_std/no_std_test/src/main.rs
+++ b/no_std/no_std_test/src/main.rs
@@ -47,3 +47,6 @@ extern "C" fn rust_eh_register_frames() {}
 
 #[no_mangle]
 extern "C" fn rust_eh_unregister_frames() {}
+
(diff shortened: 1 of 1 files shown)


schungx:
Can you push the PR to my dev repo: https://github.com/schungx/rhai

It makes it easier for me to merge this into the current stream.

schungx:
Maybe we shouldn't eliminate `rust_eh_unregister_frames`... not sure if it'll be used elsewhere  somewhere...

How about just adding `_Unwind_Resume`?

Friz64:
> How about just adding `_Unwind_Resume`?

This is exactly what this PR is doing, no?

Friz64:
> Can you push the PR to my dev repo

https://github.com/schungx/rhai/pull/63

Issue #323, opened 2021-01-05
rhai_repl: add readline support.


stustd:
I miss readline support in the REPL: scrolling through statement history is very convenient.

schungx:
What do you mean?  Something like `!123` to recall certain lines?

Right now can you use the arrow keys to scroll through past lines?  I can do this with Windows, not very sure with bash...

schungx:
Alright, I tried it with `bash` on Linux and it doesn't seem to recognize the arrow keys.  Not sure how I can add this feature, because I'm just reading from `stdin`...  any pointers?

stustd:
[This](https://lib.rs/crates/rustyline) probably will do it.

schungx:
`rhai-repl` is intended to be an example only for tutorial purposes.  Right now it is not constructed as a stand-alone tool.

In the future, `rhai-repl` and `rhai-run` may be promoted to full-blown tools, and they'll probably need to handle options etc.

In the meantime, I believe you can easily add `rustyline` to `repl` yourself...

stustd:
That's true, I "can [..and might] easily add rustyline to repl" myself, but since they're in the master's `src/bin` directory I guessed they'd be promoted sooner than later.

schungx:
I would hesitate to pull in another crate just for this tool... Maybe add new `[[bin]]` sections to `Cargo.toml` and put `rustyline` under the `rhai-repl` bin...

Would you be nice enough to revise these two tools to make them proper?  *hint*  *hint*  *hint*  *wink* *wink* *knudge* *knudge* say no more...

stustd:
> I would hesitate to pull in another crate just for this tool... Maybe add new `[[bin]]` sections to `Cargo.toml` and put `rustyline` under the `rhai-repl` bin...

That would be the way.
> 
> Would you be nice enough to revise these two tools to make them proper? _hint_ _hint_ _hint_ _wink_ _wink_ _knudge_ _knudge_ say no more...

If I had the time I would... (sorry to disappoint you...). However, I'm pretty familiar with `clap` (options parsing) and can quickly put this functionality together if I you can provide me with input what a rhai repl needs.

That said, at the moment my `rhai` work is focused on getting imported scripts compiled in a `rhai` top-level script;  [I followed your advise](https://discord.com/channels/767611025456889857/767611025456889860/796119258865729556) and it works. Is there a way to walk the AST so that I can retrieve all imports in a top-level script (this would avoid a separate `imports.rhai` script).

[...]

PR #516, opened 2022-01-29
poc: adds rustyline to rhai-repl

jonathanstrong:
saw discussion about this (https://github.com/rhaiscript/rhai/issues/323) but did not understand what the hurdle was to adding rustyline as an optional dep. perhaps I am missing something but I believe this pull request does accomplish that goal - rustyline only needed to build the repl. to me the ability to navigate command history in a repl is a huge deal!

this is a quickie, poc implementation of this to gauge interest. one known shortcoming is it breaks the use of `\` to continue the code onto the next line. also - not sure if you want to be saving the repl history in the local directory like I did. but let me know what you think!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ Cargo.lock
 benches/results
 before*
 after*
+.rhai-repl-history.txt
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,12 +15,22 @@ include = ["**/*.rs", "scripts/*.rhai", "**/*.md", "Cargo.toml"]
 keywords = ["scripting", "scripting-engine", "scripting-language", "embedded"]
 categories = ["no-std", "embedded", "wasm", "parser-implementations"]
 
+[lib]
+name = "rhai"
+path = "src/lib.rs"
+
+[[bin]]
+name = "rhai-repl"
+path = "src/bin/rhai-repl.rs"
+required-features = ["rustyline"]
+
 [dependencies]
 smallvec = { version = "1.7", default-features = false, features = ["union", "const_new" ] }
 ahash = { version = "0.7", default-features = false }
 num-traits = { version = "0.2", default-features = false }
 smartstring = { version = "0.2.8", default-features = false }
 rhai_codegen = { version = "1.2", path = "codegen", default-features = false }
+rustyline = { versoin = "9", optional = true }
 
 [dev-dependencies]
 serde_bytes = "0.11"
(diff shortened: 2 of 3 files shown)


schungx:
This looks great!  Just a couple of comments:

> one known shortcoming is it breaks the use of `\` to continue the code onto the next line. also - not sure if you want to be saving the repl history in the local directory like I did. but let me know what you think!

I happen to believe the `'` continuation is important, because I use it all the time to let `repl` compile a few lines as a single script... without it, I'd need to concat everything into a single line which is tough for debugging as well.

When I have a script that I want to copy to `repl` for testing, I just globally add `\` at the end of each line, paste, then it is OK.  Without it would be a major loss.

Secondly, I'm not sure it is wise to add an optional dependency to Rhai just to put this feature into `repl`.  We probably should start thinking splitting `rhai-repl` into a separate crate with its own dependencies.  In which case we can also use things like `clap` to have proper command line arguments...

schungx:
Merging into a branch... let me experiment to see if adding `\` will work well...

schungx:
Hhmmm....  it seems to work just fine (on Windows anyway), with `\` line continuation!  Even better, it even puts in the entire multi-line script when I recall the history instead of line by line!

schungx:
I essentially have this... not much different from yours:

```rust
    'main_loop: loop {
        input.clear();

        loop {
            match rl.readline("rhai-repl> ") {
                // Line continuation
                Ok(line) if line.ends_with("\\") => {
                    input += &line[..line.len() - 1];
                    input.push('\n');
                }
                Ok(line) => {
                    input += &line;
                    if !input.is_empty() {
                        rl.add_history_entry(input.clone());
                    }
                    break;
                }

                Err(ReadlineError::Interrupted) | Err(ReadlineError::Eof) => break 'main_loop,

                Err(err) => {
                    eprintln!("Error: {:?}", err);
                    break 'main_loop;
                }
            }
        }
```

schungx:
I see only one problem... `rustyline` is optional, so `rhai-repl` won't be built automatically when the user issues only `cargo build`.

schungx:
By mapping <kbd>Ctrl</kbd>-<kbd>Enter</kbd> or <kbd>Shift</kbd>+<kbd>Enter</kbd> to `Cmd::Newline` I can bypass the need to use `\` as continuation character!

That's because I can now enter newlines directly!

schungx:
Mapping <kbd>Shift</kbd><kbd>Enter</kbd> to `Cmd::Newline` allows me to input an entire script without using the `\` continuation character!

However, one thing is... I cannot put a prompt on each new line, therefore the lines won't match up nicely.

schungx:
#517  has `rustyline` in `rhai-repl`.

Thanks for your POC implementation!

PR #540, opened 2022-03-17
codegen: Fix hygiene mismatch between function `args` uses and definitions

petrochenkov:
Fixes rhai proc macros to work with rustc changes from https://github.com/rust-lang/rust/pull/92472.

Without this patch definitions and uses of `args` may be generated with different hygienic contexts causing name resolution errors.
One context is `Span::call_site()` used by default by `quote!`, and another is inherited from `arg_type` and used by `quote_spanned!(arg_type.span() => ...)`.
This PR changes context for all `args` to `Span::call_site()`, so they can always be resolved.

### DIFF
--- a/codegen/src/function.rs
+++ b/codegen/src/function.rs
@@ -774,14 +774,14 @@ impl ExportedFn {
                         syn::Type::Path(ref p) if p.path == string_type_path => {
                             is_string = true;
                             is_ref = false;
-                            quote_spanned!(arg_type.span() =>
+                            quote_spanned!(arg_type.span().resolved_at(Span::call_site()) =>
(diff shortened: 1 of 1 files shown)


schungx:
I don't really know too much about proc macros, so I'd take your word for it!

BTW, do we have to change any other places?  Or just these two?

petrochenkov:
This is the change that is enough to be compatible with https://github.com/rust-lang/rust/pull/92472, I'm not sure about other places.
The macros implemented in function.rs are not for public use, as I understand, so if there are any other hygiene mismatches they won't be observable anyway until these macros are used in some new way.

schungx:
Alright.  Merge it is!

PR #566, opened 2022-05-23
use smartstring's deserializer to support non borrowed strings

Geal:
The map visitor for Dynamic was expecting a &str for the key, but the
serde_json deserializer internally uses a Cow string, which can be
Borrowed or Owned. In the case of Owned, the serde_json key deserializer
is calling visit_string on the Visitor, which for &str will result in
the error:
Error("invalid type: string \"a\", expected a borrowed string", line: 0,
column: 0)

smartstring actually has its own Visitor implementation that handles
both cases, so we can use it instead of an explicit conversion.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ rustyline = { version = "9", optional = true }
 
 [dev-dependencies]
 serde_bytes = "0.11"
+serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
 
 [features]
 default = []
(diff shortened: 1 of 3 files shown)


schungx:
Great!  The serde implementation was done before `SmartString` was used, so I'll just merge.

Geal:
one of the builds fails on this, I don't really see why https://github.com/rhaiscript/rhai/runs/6557639791?check_suite_focus=true

Geal:
ah, maybe the test should have the attribute `#[cfg(not(feature = "no_object"))]`

schungx:
Fixed.

PR #584, opened 2022-07-20
fix: optimize_stmt should use fn `position`

quake:


quake:
I'm trying to integrate rhai into one of my projects, which requires custom parsing struct, so I started reading the source code of rhai, and while reading your recent PRs, I think it should be a bug here (but I don't know how to write a unit test to reproduce it)

schungx:
Yes, you're right.   The optimizer is a bit difficult to test, especially when it is the `switch` statement we're talking about, as the hashing function is not deterministic (it varies from run to run).

Usually, the way I test the optimizer is to just dump the AST as text and compare the output...

PR #590, opened 2022-07-25
feat: Basic definition files

tamasfe:
Initial definition file generation to be used with the LSP. There is no support for nested modules yet, as it's not implemented in rhai-lsp.

I had no idea what to do with the types really, sometimes they are very rusty including complex generics and paths with `crate::Something` which makes some of them unusable. The generated types look somewhat uniform now for the core library. I decided to keep primitives lower-case, since everyone is used to things like `i64` as opposed to `I64`.


tamasfe:
Ah, I used `std::io` which is probably not always. I'm not sure how to require the availability of `std` as it's currently not an opt-in feature` but an opt-out one which makes handling it a bit awkward. Thoughts on this?

schungx:
> Ah, I used `std::io` which is not always available. I'm not sure how to require the availability of `std` as it's currently not an opt-in feature` but an opt-out one which makes handling it a bit awkward. Thoughts on this?

Usually I'd just put `#[cfg(not(feature = "no_std"))]` and WASM on the whole thing.  `std::io` is not available for `no-std` and WASM mainly.

`src/api/files.rc` has the following:

```rust
#![cfg(not(feature = "no_std"))]
#![cfg(not(target_family = "wasm"))]
```

which excludes the entire file.

Cargo.toml
```diff
@@ -82,6 +82,10 @@ name = "rhai-run"
 name = "rhai-dbg"
 required-features = ["debugging"]
 
+[[example]]
```
schungx, line 85: I remember that, in Rust, if you put one `[[example]]`, you need to list all examples and it'd throw away the auto-generated list.

I'm not very sure though, but this can easily be done by putting a feature gate on the example file.  Quite a few examples which depend on features are done this way.

This would also make all the tests pass regardless of features.

tamasfe:
After a second thought, I only need `io` for one function only, I can put a condition on that, and it simply won't be available under `no_std`.

examples/definitions/main.rs
```diff
@@ -0,0 +1,40 @@
...
+#[export_module]
+pub mod general_kenobi {
+    /// Returns a string where `hello there `
+    /// is repeated `n` times.
+    pub fn hello_there(n: i64) -> String {
+        use std::convert::TryInto;
+        "hello there ".repeat(n.try_into().unwrap())
+    }
+}
+
```
schungx, line 12:
```rust
#[cfg(feature = "metadata")]
```

examples/definitions/main.rs
```diff
@@ -0,0 +1,40 @@
...
+hello_there = general_kenobi::hello_there(4 minus 2);
+"#,
+        )
+        .unwrap();
+
+    engine
+        .definitions_with_scope(&scope)
+        .write_to_dir("examples/definitions/.rhai/definitions")
+        .unwrap();
+}
```
schungx, line 40: At end:

```rust

#[cfg(not(feature = "metadata"))]
fn main() {
    panic!("This example requires 'metadata'.");
}
```

src/definitions.rs
```diff
@@ -0,0 +1,306 @@
...
+                writer.write_str(", ")?;
+            }
+            first = false;
+
+            let (param_name, param_type) = self
+                .metadata
+                .params_info
+                .get(i)
+                .map(|s| {
+                    let mut s = s.split(':');
```
schungx, line 229: You might want to do `s.splitn(':', 2);` here, otherwise:  `x : Range<path::to::my::type>` gets also split via the `::` path separators.

schungx (changes requested):

schungx:
[builtin.d.txt](https://github.com/rhaiscript/rhai/files/9186882/builtin.d.txt)

These are definitions for the built-in operators.

Cargo.toml
tamasfe: Does not seem to be the case, at least in 1.61, the examples seem to work fine.

src/definitions.rs
```diff
@@ -0,0 +1,306 @@
...
+        .unwrap_or(ty);
+
+    // FIXME(tamasfe): some of the given types are unusable,
+    // e.g. "Range<crate" is the entire type this function receives,
+    // same with "crate", I have no idea where these come from.
+    ty.replace("Range<crate", "?")
+        .replace("crate", "?")
+        .replace("Iterator<Item=", "Iterator<")
+        .replace("Dynamic", "?")
+        .replace("INT", "int")
```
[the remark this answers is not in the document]
tamasfe: Thanks, unfortunately it's not enough, I still need to do some plumbing afterwards.

src/definitions.rs
tamasfe: Ah yeah that was it, thanks.

Cargo.toml
schungx: Hhmmm.... I remember something with breaking CI with different features if I use `required-features`.  Let me check that out...

src/definitions.rs
```diff
@@ -0,0 +1,319 @@
...
+
+        fs::create_dir_all(path)?;
+
+        fs::write(path.join("__static__.d.rhai"), self.static_module())?;
+
+        if self.scope.is_some() {
+            fs::write(path.join("__scope__.d.rhai"), self.scope())?;
+        }
+
+        for (name, decl) in self.modules() {
```
schungx, line 95: `#[cfg(not(feature = "no_module"))]`

src/definitions.rs
```diff
@@ -0,0 +1,319 @@
...
+            if !first {
+                writer.write_str("\n\n")?;
+            }
+            first = false;
+
+            if f.metadata.access == FnAccess::Private {
+                continue;
+            }
+
+            f.write_definition(
```
schungx, line 196: `#[cfg(not(feature = "no_custom_syntax"))]`

src/definitions.rs
```diff
@@ -0,0 +1,319 @@
+use crate::{
+    module::FuncInfo, plugin::*, tokenizer::is_valid_function_name, Engine, Module, Scope,
+};
+use core::fmt;
+
+#[cfg(feature = "no_std")]
```
schungx, line 6: Rhai uses the dependent crate `no-std-compat` so you can just add:

```rust
#[cfg(feature = "no_std")]
use std::prelude::v1::*;
```

on top of your file and there is no need to manually import anything in the std prelude.

src/definitions.rs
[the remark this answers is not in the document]
schungx: Beware, though, that some people may register functions with `i64`, say, instead of `INT`, so the signature will not have `int` there.  Perhaps you want to also map `std::any::type_name::<crate::INT>()` and `std::any::type_name::<crate::FLOAT>()` back to `int` and `float` to be consistent.

Cargo.toml
tamasfe: The CI seems to pass, there are only some issues where I use feature-gated items.

schungx: I checked. It is OK!  So go ahead!

schungx: BTW, I remember now.  There is `required-features` which works, but no `cannot-have-features` so for most of my examples I'd need to put the feature gate in the file itself.

schungx:
BTW, I'd suggest putting `definitions.rs` under `api` since it is a public API.

schungx:
I also suggest you put an `iter_files` method to `Definition`, which would iterate a filename-content pair of strings.

`write_to_dir` can simply be implemented on top of this.

Other users can use this stream to, say, store all the definition files into a `zip` archive, for example.  In general do whatever they want to do with it.

schungx:
I just loaded the definitions file and it removed most of the errors flagged by the LSP!  Bravo!

src/definitions.rs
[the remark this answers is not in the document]
tamasfe: Last time I checked even `i64` and `i32` were incompatible in Rhai (as in no implicit cast), so I guess it's fine to make these types distinct.

tamasfe:
> I just loaded the definitions file and it removed most of the errors flagged by the LSP! Bravo!

Just wait until you see the new HIR panics I've just introduced to make the lsp unusable again.

> In fact, any thoughts of packing a bunch of definition files into a zip?

Eventually, probably. It'll bring in quite a few amount of dependencies, also it's not trivial how to use it in an LSP, as we'd have to unpack it somewhere anyway.

tamasfe:
> I modified the built-in definition file (attached) and now it parses fine.

I did as well, decided to separate the operators from the `contains` function which I merged into the hand-written builtin definition.

I think this should be ready to merge on my part if the CI finally passes and you approve.

schungx:
Once the tests pass, I think this is good to merge.

Just one thing: `expr1 in expr2` compiles to `contains(expr2, expr1)`.  The definition files only have definitions for `contains`, but not the `in` operator.  Should we automatically generate an `in` version whenever we encounter a `contains`?

`contains` is a constant exposed via `OP_CONTAINS`, so you don't have to hard code it.

schungx (approved):

schungx:
Now I'll start to write the docs for definitions in the book!

tamasfe:
Is `in` just a sugar for `contains` calls? If yes, we should generate the operator (or write it by hand) whenever we encounter `contains`, yes.

schungx:
> Is `in` just a sugar for `contains` calls? If yes, we should generate the operator (or write it by hand) whenever we encounter `contains`, yes.

Yes, it is just sugar.  It compiles directly to a call to `contains`.

PR #596, opened 2022-07-27
Definitions: nested modules, single file, refactors

tamasfe:
- implemented nested modules with the `module foo { ... }` syntax
- exposed builtin function and operators definitions as separate functions
- added some basic configuration for generation and made some functions internally reconfigurable; the user now can choose whether we should include `module ...` at the beginning of the definitions, but we always need those when writing to files.
- with nested modules we can actually put everything under `module static;`, so I exposed this feature as well.
- refactored so that `write_to_dir` actually uses `iter_files`, I seem to have forgotten that in the previous PR...

I did not really bother with indentation and formatting, I made sure that everything is readable, but beyond that we'll have to run the generated content through a formatter eventually.

Also I'd probably not include this code in releases or at least with a note that it's not semver-compatible. I imagine we might have a quite a few breaking changes as it evolves alongside the LSP.

### DIFF
--- a/examples/definitions/.rhai/all_in_one.d.rhai
+++ b/examples/definitions/.rhai/all_in_one.d.rhai
@@ -1,0 +1,6256 @@
+module static;
+
+op ==(int, int) -> bool;
+op !=(int, int) -> bool;
+op >(int, int) -> bool;
+op >=(int, int) -> bool;
+op <(int, int) -> bool;
+op <=(int, int) -> bool;
+op &(int, int) -> int;
+op |(int, int) -> int;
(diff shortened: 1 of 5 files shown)


schungx:
> Also I'd probably not include this code in releases or at least with a note that it's not semver-compatible. I imagine we might have a quite a few breaking changes as it evolves alongside the LSP.

We can probably put this under `internals` for the time being... anything `internals` is supposed to be volatile.

schungx:
https://github.com/rhaiscript/rhai/pull/597  gates definitions under the `internals` feature.

PR #601, opened 2022-07-29
Add include_standard_packages to Definitions.

schungx:


schungx:
@tamasfe   This adds an option `include_standard_packages` to `Definitions`.  It allows the user to omit functions from the standard builtin libraries.

With this, I can merge the `gen_fn_metadata_to_json` API with `Definitions` and then deprecate the original API.

PR #609, opened 2022-08-08
add a trait to register custom type and it's methods, getters, setters, etc.

tguichaoua:
## What this PR do
Add a trait `RhaiCustomType` with a method to build the type i.e. register the type itself, it's methods, getters, setters, etc...

Like [mlua's UserData](https://docs.rs/mlua/latest/mlua/trait.UserData.html) trait.

## Why

Make the code clearer :
- Where the engine is created : only one line for each custom type `engine.build_type::<MyType>()`
- The code to build the type is next the type itself

## Notes

- The concept is named "build" a type to avoid confusion with the concept of "register" a type.
- The trait (`RhaiCustomType`) and it's method (`build`) could have a better name.

## Example
```rust
use rhai::{Engine, RhaiCustomType};

#[derive(Debug, Clone, Eq, PartialEq)]
struct TestStruct {
    field: i64
}

impl TestStruct {
    fn new() -> Self {
        Self { field: 1 }
    }
    fn update(&mut self, offset: i64) {
        self.field += offset;
    }
}

impl RhaiCustomType for TestStruct {
    fn build(engine: &mut Engine) {
        engine
            .register_type::<Self>()
            .register_fn("new_ts", Self::new)
            .register_fn("update", Self::update);
    }
}

fn main() -> Result<(), Box<rhai::EvalAltResult>> {

    // Register API for the custom type.
    engine.build_type::<TestStruct>();

    assert_eq!(
        engine.eval::<TestStruct>("let x = new_ts(); x.update(41); x")?,
        TestStruct { field: 42 }
    );

    Ok(())
}
```

### DIFF
--- a/src/api/build_type.rs
+++ b/src/api/build_type.rs
@@ -56,7 +56,7 @@ impl Engine {
     /// Build a custom type for use with the [`Engine`].
     /// i.e. register the type, getters, setters, methods, etc...
     ///
-    /// See [`RhaiCustomType`].
+    /// See [`CustomType`].
     pub fn build_type<T>(&mut self) -> &mut Self
     where
         T: CustomType,
--- a/src/api/build_type.rs
+++ b/src/api/build_type.rs
@@ -6,7 +6,7 @@ use crate::Engine;
 /// # Example
 ///
 /// ```
-/// use rhai::{Engine, RhaiCustomType};
+/// use rhai::{Engine, CustomType};
 ///
 /// #[derive(Debug, Clone, Eq, PartialEq)]
 /// struct TestStruct {
@@ -22,7 +22,7 @@ use crate::Engine;
 ///     }
 /// }
 ///
-/// impl RhaiCustomType for TestStruct {
+/// impl CustomType for TestStruct {
 ///     fn build(engine: &mut Engine) {
 ///         engine
 ///             .register_type::<Self>()
--- a/src/api/build_type.rs
+++ b/src/api/build_type.rs
@@ -1,12 +1,17 @@
-use crate::Engine;
+use core::marker::PhantomData;
+
+use crate::{
+    func::SendSync, types::dynamic::Variant, Engine, Identifier, RegisterNativeFunction,
+    RhaiResultOf,
+};
 
 /// Trait to build a custom type for use with the [`Engine`].
-/// i.e. register the type, getters, setters, methods, etc...
+/// i.e. register the type and its getters, setters, methods, etc...
 ///
 /// # Example
 ///
 /// ```
-/// use rhai::{Engine, CustomType};
+/// use rhai::{CustomType, TypeBuilder, Engine};
 ///
 /// #[derive(Debug, Clone, Eq, PartialEq)]
 /// struct TestStruct {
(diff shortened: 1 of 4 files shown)


src/api/build_type.rs
```diff
@@ -0,0 +1,67 @@
...
+///
+/// # #[cfg(not(feature = "no_object"))]
+/// assert_eq!(
+///     engine.eval::<TestStruct>("let x = new_ts(); x.update(41); x")?,
+///     TestStruct { field: 42 }
+/// );
+/// # Ok(())
+/// # }
+/// ```
+pub trait RhaiCustomType {
```
schungx, line 49: I would suggest calling it just `CustomType` since `rhai::CustomType` is already quite unambiguous.

src/api/build_type.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// Builds the custom type for use with the [`Engine`].
+    /// i.e. register the type, getters, setters, methods, etc...
+    fn build(engine: &mut Engine);
+}
```
schungx, line 53: Let's spend a little time thinking about what other functions may be needed.  Once this goes out, we can no longer easily change this trait.

schungx:
Also care to add a test?

src/api/build_type.rs
tguichaoua: Yes, I think it's wise.

Also I thought of not expose directly `Engine` to avoid potiential side effects and using a wrapper that only expose methods related to building type.

Here's just a prototype I have in mind
```rust
struct TypeBuilder<'a, T> {
    engine: &'a mut Engine,
    name: Option<&'static str>,
    _marker: PhantomData<T>,
};

impl<'a, T> TypeBuilder<'a> {
    fn with_name(&mut self, name: &'static str) -> &mut self {
        self.name = Some(name);
        self
    }

    fn add_method<N, F>(&mut self, name: N, method: F) -> &mut self 
    where /* omited */
    {
        self.engine.register_fn(name, method);
        self
    }

    // other method to add getters, setters, etc.

    fn build(self) {
        /* empty */
    }
}

impl<'a, T> Drop for TypeBuilder<'a, T> {
    fn drop(&mut self) {
        if let Some(name) = self.name {
            self.engine.register_type_with_name::<T>(name);
        } else {
            self.engine.register_type::<T>();
        }
    }
}
```

src/api/build_type.rs
```diff
@@ -0,0 +1,67 @@
...
+
+impl Engine {
+    /// Build a custom type for use with the [`Engine`].
+    /// i.e. register the type, getters, setters, methods, etc...
+    ///
+    /// See [`RhaiCustomType`].
```
schungx, line 59: Should fix to `CustomType`.

src/api/build_type.rs
schungx: I have the feeling that `engine.build_type::<Type>()` looks the easiest...

tguichaoua: I implement the prototype on another branch.
Here's what I have in mind: https://github.com/tguichaoua/rhai/blob/4432d5063a44c082d1f20c4c4889865b927ed7d9/tests/build_type.rs

schungx: This actually looks quite interesting!  I would suggest go with this one instead!

Also, why not mark it (and all the related types) as `volatile` (usually I use a specially-phrased `#[deprecated]` attribute  to do that) so that users know that this API might change.

schungx:
I think the style in https://github.com/tguichaoua/rhai/blob/4432d5063a44c082d1f20c4c4889865b927ed7d9/tests/build_type.rs is best for the user, and it gives expansion possibilities later on without changing the API.

src/api/build_type.rs
tguichaoua: Why not use a feature like `type_builder` and state in the documentation that this feature is unstable and may change in the future ?
Once stabilized, we can remove the `#[cfg(feature = "type_builder)]` and in the documentation say that the `type_builder` feature does nothing and will be removed in the next major update.

schungx: Well, we would actually want to encourage the user to use it without worrying about changing code later on.

If we force the user to pull in a feature, then once that feature goes away, it will be a breaking change.

If we simply mark it as volatile, however, the user can simply use it, and as long as we don't change the API, there is nothing he needs to do when we  make it non-volatile - the warning just disappears.

tguichaoua: Ok

PR #625, opened 2022-08-23
Empty for Arrays

ltabis:
Would it be interesting to have an `empty` getter on arrays to check if they're empty instead of using `array.len == 0` ?

### DIFF
--- a/examples/definitions/.rhai/all_in_one.d.rhai
+++ b/examples/definitions/.rhai/all_in_one.d.rhai
@@ -3047,6 +3047,9 @@ fn keys(map: Map) -> Array;
 /// Number of elements in the array.
 fn len(array: Array) -> int;
 
+/// Return true if the array is empty.
+fn is_empty(array: Array) -> bool;
+
 /// Return the length of the BLOB.
 ///
 /// # Example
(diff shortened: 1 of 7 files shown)


schungx:
It probably should be called `is_empty` similar to Rust...

But I hesitate to add such a function as we'd need to add it for all other data types to be consistent...

schungx:
It probably should be called `is_empty` similar to Rust...

But I hesitate to add such a function as we'd need to add it for all other data types to be consistent...

ltabis:
> But I hesitate to add such a function as we'd need to add it for all other data types to be consistent...

Except for `map` and `string`, which other type would need the `is_empty` function ?

schungx:
Probably BLOBs, ranges, ...

ltabis:
I've added `is_empty` to map, blobs and strings. For ranges, I'm not sure if it's necessary since they don't expose `len` themselves.

schungx:
Well they dont expose `len` but they do check for containment so I think it probably makes sense.

ltabis:
If you are talking about the `RangeCase`, `is_empty` is already implemented just not exposed in rhai, is that ok ?

schungx:
Sorry, I didn't make myself clear.

Methods on ranges are defined in `iter_basic.rs` under `mod range_functions`.

ltabis:
Woops it's my bad, should I add `len` too ?

ltabis:
I've updated it!

PR #653, opened 2022-10-11
Ensure Wasm build works

### LINKED ISSUE #652: Question: Proper way to make a WASM build
Hi, I was very excited to here this could run in wasm and wanted to test it out, first to see relative byte code sizes I could expect before embedding it in my program. However, I had a number of errors compiling (on Rust 1.62). The minimal version that finally worked was:

`cargo build --release --target wasm32-unknown-unknown --no-default-features`
But I think I really would want:
`cargo build --release --target wasm32-unknown-unknown --no-default-features --features wasm-bindgen`

The following variants all errored:

* `cargo build --release --target wasm32-unknown-unknown`
* `cargo build --release --target wasm32-unknown-unknown --features wasm-bindgen`

```
   Compiling getrandom v0.2.7
error: the wasm32-unknown-unknown target is not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support
```

* `cargo build --release --target wasm32-unknown-unknown --features wasm-bindgen,no_std --no-default-features` also errored:

```
error: `wasm-bindgen` cannot be used with `no-std`
   --> src/lib.rs:447:1
    |
447 | compile_error!("`wasm-bindgen` cannot be used with `no-std`");
```

It seems funny to disable both `std` and `no_std` to get it to compile. Is this the proper approach?

NB: I looked at [the CI Scripts](https://github.com/rhaiscript/rhai/blob/main/.github/workflows/build.yml) to see how you build wasm there, and it is not included. May be nice to add a small section there, to at least test it compiles to wasm properly (even if it doesn't spin up node server and run the wasm)

ethanfrey:
Closes #652

- [x] Add Wasm build step to CI
- [x] Fix feature flags so it compiles on Wasm

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,6 +110,32 @@ jobs:
       with:
         command: build
         args: --manifest-path=no_std/no_std_test/Cargo.toml ${{matrix.flags}}
+  wasm:
+    name: Check Wasm build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        flags:
+          - ""
+          - "--no-default-features"
+          - "--features wasm-bindgen"
+          - "--no-default-features --features wasm-bindgen"
+      fail-fast: false
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ env.RUST_MSRV }}
+          override: true
+          target: wasm32-unknown-unknown
+      - name: Build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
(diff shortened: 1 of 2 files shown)


schungx:
Thanks for the PR. I'll need yo go thru it more carefully as I see quite extensive changes.

For features, I realize that `ahash/compile-time-rng` can be set always because it is always trumped by `ahash/runtime-rng`, at least in the current source code. So a simpler solution is to pull `ahash/compile-time-rng` out of `std`.

ethanfrey:
Okay, I made https://github.com/rhaiscript/rhai/pull/654

ethanfrey:
Also, created https://github.com/rhaiscript/rhai/pull/655 for the features changes more or less how you suggested.

ethanfrey:
Closing this as deprecated by the two PRs linked above

PR #655, opened 2022-10-11
Wasm friendly features

ethanfrey:
Builds on #654 (review and merge that first)

Updates features so they work for wasm out of the box, and can opt-in to `runtime-rng`.

Improve CI matrix for testing.


schungx:
I hesitate to add too many new features as there are already a ton.  Too many new features catering only for wasm is probably not a good idea.

I think I'd settle on setting `ahash/compile-time-rng` and pulling `ahash/runtime-rng` out of `std` into `default` (which trumps `compile-time-rng`).  This way it is rather clean and `--features std,wasm-bindgen --no-default-features` work.

In other words:

* default features = opt into `runtime-rng`
* no default features + `std` = `compile-time-rng` plus `std` stuff (which doesn't do much right now)
* no default features = `compile-time-rng` plus no `std` stuff

As for `getrandom/js` etc. I'd move that into the Book as documentation.

schungx:
For example, I just checked (and put into CI): `wasm32-wasi` works as-is without any modification.

ethanfrey:
Okay, seems like a reasonable approach.

Are you going to make a pr for that approach? If so, feel free to close this.

PR #657, opened 2022-10-14
Make time references an optional import

ethanfrey:
When I compile this to wasm with the new format

`cargo build --target wasm32-unknown-unknown --no-default-features --features std`

It works properly, but it creates an `env.now` import, to load system time. I am not running in a browser, but in another embedded environment, that doesn't have access to clock time (or random number generators, already removed). So, the final *.wasm output doesn't run properly ("unsupported import").

I noticed this time dependency was conditionally disabled for `no_std`, but `no_std` is incompatible with `wasm32-unknown-unknown` builds. 

My decision was to change the conditions for time to a new `no_time` feature (like many of the other `no_*` features) and then include `no_time` when `no_std` is defined. This allows any "normal" `std` build also disable the system time dependency if desired.

I know it does add one more feature flag, but it is strictly optional and the features involved we already conditionally compiled, just I moved to a different flag.

Using this patch, the resulting code runs properly in my environment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
 [features]
 default = ["std", "ahash/runtime-rng"]  # ahash/runtime-rng trumps ahash/compile-time-rng
 std = ["ahash/std", "num-traits/std", "smartstring/std"]
+no_time = []                    # remove all time-related imports
 unchecked = []                  # unchecked arithmetic
 sync = []                       # restrict to only types that implement Send + Sync
 no_position = []                # do not track position in the parser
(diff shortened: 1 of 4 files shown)


schungx:
Actually you need to specify the `wasm-bindgen` feature to have `instant` pull in the correct code.

ethanfrey:
We actually don't support wasm-bindgen, as this is not a browser setting. I am running inside [wasmer](https://wasmer.io) with minimal imports (heavily sandboxed)

`Wasm contract requires unsupported import: "__wbindgen_placeholder__.__wbindgen_object_drop_ref" ...`

I got this both when adding the `wasm-bindgen` dependency directly on `rhai`, as well as adding `instant = { version = "0.1.10", features = ["wasm-bindgen"] }` to my dependencies.

schungx:
Ah I see. So you need a eay to disable the time features.

schungx:
But I probably would also remove all the time-related code from `Dynamic` etc.

Although for some reason those work under wasm. I believe it is the getting the current time that is tripping up wasm.

schungx:
@ethanfrey   I'll close this PR and put in one where I remove all occasions of `TimeStamp` when `no_time` is used.  This should make the build smaller by excluding all time-related features.

schungx:
@ethanfrey `no_time` has landed. Can you test it out. Thx.

ethanfrey:
Wow! Great job.

I just tested out rev `60550e1` on the tip of main branch, and it works great, with `default_features = false, features = ["std", "no_time"]`. Compiles without problem, and the resultant wasm doesn't have any unrecognized imports.

I very much appreciate your responsiveness and thank you for unblocking for my use case. I hope to share a final product in a few weeks.

PR #659, opened 2022-10-15
Add no_time feature to disable timestamps.

schungx:


ethanfrey:
Amazing work! Thank you so much for this!

PR #662, opened 2022-10-31
New Static Hash Feature

l1npengtul:
`stable_hash` is a feature that allows users to set their own hash. It is nightly only (once_cell), and incompatible with `no_std`.

This introduces a new API for setting and getting the Rhai AHash.
```rust
// config.rs
pub fn set_rhai_ahash_seed(new_seed: Option<[u64; 4]>) -> Result<(), Option<[u64; 4]>>;

pub fn get_rhai_ahash_seed() -> Option<[u64; 4]>;
```

Please check the included documentation to learn more.

Changelog:

- Make `config.rs` public
    -  Write associated Docs
- New AHash API:
    -  `set_rhai_ahash_seed(new_seed: Option<[u64; 4]>) -> Result<(), Option<[u64; 4]>> `: Sets the Rhai Ahash seed. This can only be called once per program. (`stable_hash` only) 
    -  `get_rhai_ahash_seed() -> Option<[u64; 4]>`: Gets the Rhai Ahash seed. None if not set
- `build.rs`: template is now feature gated, if the user has `stable_hash` enabled, template will not be done.
- `build.template`: Updated to reflect new API
- `build.yml`: Add nightly matrix to CI nightly only features (aka `stable_hash`)
- `Cargo.toml`: Add `stable_hash` feature with dependency `std`

Considerations:

- We can make this `no_std` compatible by rolling our own `OnceLock` by using `SyncUnsafeCell`. This however, would require unsafe code.


.gitignore
```diff
@@ -8,3 +8,6 @@ Rhai.toml
 **/*.bat
 doc/rhai-sync.json
 doc/rhai.json
+.idea/
```
schungx, line 11: Probably should remove `.idea` here as they don't exist in the official repo.

Cargo.toml
```diff
@@ -63,6 +63,7 @@ unicode-xid-ident = ["unicode-xid"] # allow Unicode Standard Annex #31 for ident
 metadata = ["serde", "serde_json", "rhai_codegen/metadata", "smartstring/serde"] # enable exporting functions metadata
 internals = []                  # expose internal data structures
 debugging = ["internals"]       # enable debugging
+stable_hash = ["std"]
```
schungx, line 66: Is there a way to avoid introducing _yet_ another feature flag?  Perhaps gate it under `nightly`?

build.template
```diff
@@ -1,7 +1,60 @@
...
+/// - You **MUST** call this before **ANY** Rhai operation occurs (e.g. creating an [`Engine`]).
+pub fn set_rhai_ahash_seed(new_seed: Option<[u64; 4]>) -> Result<(), Option<[u64; 4]>> {
+    AHASH_SEED.set(new_seed)
+}
+
+#[cfg(feature = "stable_hash")]
+/// Gets the current Rhai Ahash Seed.
+///
+/// See [`set_rhai_ahash_seed`] for more.
+pub fn get_rhai_ahash_seed() -> Option<[u64; 4]> {
```
schungx, line 50: The `rhai` is redundant since it is in the `rhai` crate...  better name them `get_ahash_seed()` and `set_ahash_seed()`...

.gitignore
l1npengtul: Those are intellij idea files, they are there for the same reason `.vscode` is there - IDE specific files. I think its a good idea to keep those in case someone else also uses intellij-rust (me)

Cargo.toml
l1npengtul: I don't think so - there isnt a `rustc_version` attribute unless you are willing to do build.rs trickery or pulling another crate. Keep in mind that this is incompatible with your no-std-compat crate as that one does not ship nightly features, so keeping it feature gated even in nightly is probably a good idea.

build.template
```diff
@@ -1,7 +1,60 @@
-//! Configuration settings for this Rhai build
-//!
-//! This file is auto-generated from `build.template`
+#[cfg(feature = "stable_hash")]
+use core::cell::OnceCell;
 
+// [236,800,954,213], haha funny yume nikki reference
+
+#[cfg(not(feature = "stable_hash"))]
```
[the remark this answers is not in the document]
l1npengtul: I'm having trouble understanding here, do you mean move the entire thing to a different module and `pub use` them out?

schungx:
> `stable_hash` is a feature that allows users to set their own hash. It is nightly only (once_cell), and incompatible with `no_std`.
> 
> This introduces a new API for setting and getting the Rhai AHash.
> 
> ```rust
> // config.rs
> pub fn set_rhai_ahash_seed(new_seed: Option<[u64; 4]>) -> Result<(), Option<[u64; 4]>>;
> 
> pub fn get_rhai_ahash_seed() -> Option<[u64; 4]>;
> ```
> 
> Please check the included documentation to learn more.
> 
> Changelog:
> 
>     * Make `config.rs` public
>       
>       * Write associated Docs
> 
>     * New AHash API:
>       
>       * `set_rhai_ahash_seed(new_seed: Option<[u64; 4]>) -> Result<(), Option<[u64; 4]>> `: Sets the Rhai Ahash seed. This can only be called once per program. (`stable_hash` only)
>       * `get_rhai_ahash_seed() -> Option<[u64; 4]>`: Gets the Rhai Ahash seed. None if not set
> 
>     * `build.rs`: template is now feature gated, if the user has `stable_hash` enabled, template will not be done.
> 
>     * `build.template`: Updated to reflect new API
> 
>     * `build.yml`: Add nightly matrix to CI nightly only features (aka `stable_hash`)
> 
>     * `Cargo.toml`: Add `stable_hash` feature with dependency `std`
> 
> 
> Considerations:
> 
>     * We can make this `no_std` compatible by rolling our own `OnceLock` by using `SyncUnsafeCell`. This however, would require unsafe code.

Ah, so that's why `no-std` is not supported.  Bummer, as it would have been nice.

I have a number of comments and which made into a review.

My suggestion is to leave `build.rs` there for future uses, but group the configurations into modules.  So `mod ahash` or `mod hashing` for hashing stuff.  `crate::config::hashing::get_ahash_seed()` would then do the right thing based on the build.

I would also have to change the write-up in the Book to cater for this.

l1npengtul:
> Ah, so that's why no-std is not supported. Bummer, as it would have been nice.
I can take a bit to steal the `once_cell::sync` implementation if you would need
> I would also have to change the write-up in the Book to cater for this.
I can do that if you wish.

build.template
```diff
@@ -1,7 +1,60 @@
...
 /// Set to [`None`] to disable stable hashing.
-pub const AHASH_SEED: Option<[u64; 4]> = {{AHASH_SEED}};
+///
+/// See [`set_rhai_ahash_seed`] for more.
+///
+/// You can also set this at compile time by disabling `stable_hash` and setting the `RHAI_AHASH_SEED`
+/// environment variable instead.
+///
+/// E.g. `env RHAI_AHASH_SEED ="[236,800,954,213]"`
+static AHASH_SEED: OnceCell<Option<[u64; 4]>> = OnceCell::new();
```
schungx, line 29: I wonder if there is a way to fallback to a default seed if the user does not call `set_ahash_seed`?  I guess it might not be very useful though because the compiler really needs to get this default from somewhere.

l1npengtul: Ill just add a default to `get_ahash`.

schungx:
Since you're doing this, I suppose you're using the nightly compiler for your project?

It really would be nice to have `no-std` work out... even at the expense of a little `unsafe`...

l1npengtul:
Ill change this back to a draft PR then and start stealing an implementation.

l1npengtul:
Or - I could do even more trickery and make an AtomicBool+MaybeUninit based implementation to make it work even on stable.

l1npengtul:
Renamed to static hash to avoid confusion

schungx:
A couple of things:

1) Pls delete `.idea` from `.gitignore` as they don't exist in this repo...
2) I'm wondering whether it would be a good idea to have separate templates to build separate config files.  For example, `build.template` should probably be renamed `build.hashing.template`...  and perhaps under a fold called `templates`?  That would make it possible to add future configs without having to figure out what to do with `build.template` (since right now it all goes into `src/config/hashing.rs`.

Other than this, if the CI runs, it looks cool!

schungx:
Ah, it only fails CI on formatting. Can you do a `cargo fmt`?

l1npengtul:
> 



> A couple of things:
> 
>     1. Pls delete `.idea` from `.gitignore` as they don't exist in this repo...
> 

It is there for the same reason `.vscode` is there, but if you insist...

>     2. I'm wondering whether it would be a good idea to have separate templates to build separate config files.  For example, `build.template` should probably be renamed `build.hashing.template`...  and perhaps under a fold called `templates`?  That would make it possible to add future configs without having to figure out what to do with `build.template` (since right now it all goes into `src/config/hashing.rs`.

That would probably make it more clear. 

>     3. I wonder if there is a minimal Rust version to enable `feature(once_cell)` in stable?  If so, there are probably a number of things I can do better by using `once_cell` in other parts of the code!

This doesn't rely on nightly, this works fine on stable thanks to `unsafe`.

> Ah, it only fails CI on formatting. Can you do a cargo fmt?

I'll do so once I get home :smiley:

schungx:
BTW, can we get rid of `static_hash` and just enable it by default?

In that case, can we have that:

* at _compile time_, always check for `RHAI_AHASH_SEED` and set a default seed based on this env var (or `None`)
 
* at runtime, if `rhai::config::hashing::set_ahash_seed` is called, use that fixed seed always (regardless of `RHAI_AHASH_SEED`)
 
* otherwise, at runtime, return `RHAI_AHASH_SEED` (if any)

* if no seed, then use dynamic hashing

schungx:
> It is there for the same reason `.vscode` is there, but if you insist...

Ah, I guess it is IntelliJ IDEA.  In that case leave it.

l1npengtul:
> BTW, can we get rid of `static_hash` and just enable it by default?

Ok, will work on it once I get home.

schungx:
In that case, come to think of it, that means that the `once_cell` based code can be included in `src` and moved out of `build.rs` because it would be standard code.

So that means probably `src/config/hashing_static.rs` is normal code file, while `src/config/hashing_env.rs` is generated from template by `build.rs`?

l1npengtul:
Oops, left in static hash in CI by mistake

.github/workflows/build.yml
```diff
@@ -89,13 +89,46 @@ jobs:
         command: test
         args: ${{matrix.flags}}
 
+#  nightly_features_build:
```
schungx, line 92: Can we remove this as it no longer requires nightly?  Although it is already commented out...

build.rs
```diff
@@ -18,10 +17,10 @@ fn main() {
 
     let seed = env::var("RHAI_AHASH_SEED").map_or_else(|_| "None".into(), |s| format!("Some({s})"));
 
-    contents = contents.replace("{{AHASH_SEED}}", &seed);
+    contents = contents.replace("{{ AHASH_SEED }}", &seed);
```
schungx, line 20: Requiring a space after and before the `{{ ... }}` may make the template brittle...

build.template
```diff
@@ -1,7 +1,112 @@
-//! Configuration settings for this Rhai build
+//! Fixed hashing seeds for stable hashing.
 //!
-//! This file is auto-generated from `build.template`
+//! Set to [`None`] to disable stable hashing.
+//!
+//! See [`set_rhai_ahash_seed`] for more.
+//!
+//! Alternatively, You can also set this at compile time by disabling `stable_hash` and setting the `RHAI_AHASH_SEED`
```
schungx, line 7: Should remove reference to `stable_hash`

build.template
```diff
@@ -1,7 +1,112 @@
...
+//! environment variable instead.
+//!
+//! E.g. `env RHAI_AHASH_SEED ="[236,800,954,213]"`
+// [236,800,954,213], haha funny yume nikki reference epic uboachan face numberworld nexus moment 100
```
schungx, line 11: Wonder what yume nikki reference is...  :-)

build.template
```diff
@@ -1,7 +1,112 @@
-//! Configuration settings for this Rhai build
+//! Fixed hashing seeds for stable hashing.
```
schungx, line 1: I suggest the stable hashing code moved out of `build.rs` into the source tree (e.g. `src/config/hashing_stable.rs`).

The template can keep only the env-var `AHASH_SEED` value which is written to a generated file, say `src/config/hashing_env.rs`.

build.template
```diff
@@ -1,7 +1,112 @@
...
+pub fn set_ahash_seed(new_seed: Option<[u64; 4]>) -> Result<(), Option<[u64; 4]>> {
+    AHASH_SEED.set(new_seed)
+}
+
+/// Gets the current Rhai Ahash Seed. If the seed is not yet defined, this will automatically set a seed.
+/// The default seed is not stable and may change between versions.
+///
+/// See [`set_rhai_ahash_seed`] for more.
+pub fn get_ahash_seed() -> Option<[u64; 4]> {
+    const FUNNY_YUMENIKKI_REFERENCE: Option<[u64; 4]> = {{ AHASH_SEED }};
```
schungx, line 106: This can probably simply refer to the generated `hashing_env::AHASH_SEED` instead of splicing the env var value here, so you can move this whole lot into the source tree.

schungx (commented):
In general, I believe it is better to move any non-volatile code out of `build.rs` and into the main source tree, so the build template should only contain stuff that must be done at compile time.

.github/workflows/build.yml
l1npengtul: Yep.

build.rs
l1npengtul: Will revert

build.template
l1npengtul: its 4 of the numbers that are on the ground in number world

build.rs
schungx: Unless you want it to auto-format as Rust... in that case the editor may auto-add those spaces...

schungx:
BTW, I wonder if the `SusLock` which uses atomics will works under `wasm`?

build.template
schungx: Ah, after some Googling, I got it now!

build.template
```diff
@@ -1,7 +1,112 @@
...
+//!
+//! E.g. `env RHAI_AHASH_SEED ="[236,800,954,213]"`
+// [236,800,954,213], haha funny yume nikki reference epic uboachan face numberworld nexus moment 100
+
+use core::{
+    cell::UnsafeCell,
+    marker::PhantomData,
+    mem::MaybeUninit,
+    panic::{RefUnwindSafe, UnwindSafe},
+    sync::atomic::{AtomicBool, Ordering},
```
schungx, line 18: Gate atomics under the `sync` feature?

build.template
```diff
@@ -1,7 +1,112 @@
...
+};
+
+// Safety: lol
+struct SusLock<T>
+where
+    T: 'static,
+{
+    initalized: AtomicBool,
```
schungx, line 26: Use a normal `bool` for non-`sync`?

build.template
```diff
@@ -1,7 +1,112 @@
...
+        if self.initalized.load(Ordering::SeqCst) {
+            Some(unsafe { (&*self.data.get()).assume_init_ref() })
+        } else {
+            return None;
+        }
+    }
+
+    pub fn get_or_init(&self, f: impl FnOnce() -> T) -> Option<&T> {
+        let value = f();
+        if !self.initalized.load(Ordering::SeqCst) {
```
schungx, line 50: This might contend from multiple threads...  if two threads arrive at this at the same time, then falls through at the same time, you end up setting the `data` twice (possibly with two different values).

You'd probably need test-and-set in order protect against this...

build.template
```diff
@@ -1,7 +1,112 @@
...
+            unsafe {
+                self.data.get().write(MaybeUninit::new(value));
+            }
+            self.initalized.store(true, Ordering::SeqCst);
```
schungx, line 54: I think at least move this _before_ setting `data` to minimize the time between passing the gate and then setting the gate to `true` to block future entries...

schungx:
@l1npengtul   I wonder if you're still working on this PR?  Because it looks quite OK as it stands... If it passes CI, I can probably just merge it and then do code-style cleanup later on.

l1npengtul:
Im fixing the concurrency concerns... just givr me a couple hours...

schungx:
> Im fixing the concurrency concerns... just givr me a couple hours...

No pressure at all!  Take all the time you need.

l1npengtul:
> > Im fixing the concurrency concerns... just givr me a couple hours...
> 
> No pressure at all! Take all the time you need.


I will steal 10 more concurrency primitives before this is merged /s

l1npengtul:
I fell asleep while writing this PR 💀

l1npengtul:
Yep, I think I fixed the majority of concerns.

Unless anything crops up I think this is ready for a merge :+1:

schungx:
I'll check that out once i get to my desk.

schungx:
Have you benchmarked it or checked it out with compiler explorer for the codegen?  I hope it is not adding significant overhead to get a hasher as that is done a lot...

l1npengtul:
I've run bench (well, the parts of bench that dont crash) and..

Main:
Screenshot_20221104_185749

Stable Hash:
Screenshot_20221104_185841

It doesn't seem to be that much overhead but I'll try further optimizing if needed.

l1npengtul:
Nevermind, I just needed to update cargo. Give me a moment.

l1npengtul:
Here are the benchmarks:

[rhai_bench_stablehash.txt](https://github.com/rhaiscript/rhai/files/9937144/rhai_bench_stablehash.txt)

[rhai_bench_main.txt](https://github.com/rhaiscript/rhai/files/9937145/rhai_bench_main.txt)

Overall, not much change. It could be just down to run-to-run variations.

PR #666, opened 2022-11-05
Static Hash Finetuning

l1npengtul:


schungx:
@l1npengtul  this fails on CI.  I think a few syntax errors.

l1npengtul:
im done for today... other things to work on...

ill fix it tomorrow :p

schungx:
One more nit-pick:  Function `get_or_init` always returns `Some`.  Therefore, there is really no reason for it to return `Option<T>`.  It should simply return `T` or `&T`.

src/config/hashing.rs
```diff
@@ -136,9 +136,9 @@ where
     }
 
     #[must_use]
-    pub fn get_or_init(&self, f: impl FnOnce() -> T) -> Option<T> {
-        let value = f();
+    pub fn get_or_init(&self, f: impl FnOnce() -> T) -> Option<&'static T> {
```
schungx, line 139: `get_or_init` alreays return `Some` so it should be simply `&'static T`.

l1npengtul: This can still be null since the AHASH_SEED envvar isnt always set

l1npengtul:
Ok IDK what happened with the diff...

src/config/hashing.rs
schungx: Well, `T` itself is `Option`, so if `AHASH_SEED` isn't set, it'll return `&'static None`.

schungx: No worries, I'll just merge it and fiddle with it later.

l1npengtul:
yeah

PR #669, opened 2022-11-10
Native module

ltabis:
I would like to change the visibility of the `func::native` module to public to be able to use functions like `locked_read` & `locked_write` for `rhai-dylib`'s module resolver implementation. Is a simple re-export enough or should we lock the export behind a feature gate ?


schungx:
I can make them public...  no particular harm in doing that.  Or gate them under `internals`?

They are just simple wrappers, you can simply copy them over.

schungx:
Oops, forgot about this and I just released `1.11.0`.  Oh well, we can always add this to `1.11.1`...  ;-)

ltabis:
> Oops, forgot about this and I just released `1.11.0`. Oh well, we can always add this to `1.11.1`... ;-)
I've synced the fork :)

ltabis:
@schungx done, `func::native` is public under the `internals` feature.

schungx:
Actually, I would prefer to cherry-pick the functions that are useful to be exported...  There is probably not much use to export the internal types for function signatures...

Say `pub func::native::{locked_read, locked_write};`

PR #688, opened 2022-12-30
Fix lints warnings and enhance the CI

Mathieu-Lala:
# Context

I added a job on the CI to enhance the test suite concerning the cargo feature flags. Using `cargo-hack` I caught a build error with `cargo check --no-default-features --features bin-features,no_std`, I guess more can be found.

I also fixed the linter's warnings I had when I opened the project.

Some github actions used in the CI are using old version ([`checkout@v2`](https://github.com/rhaiscript/rhai/blob/main/.github/workflows/build.yml#L79)), those can be automatically bumped using dependabot, do you wish to use this tool ?

# Summary of changes

## Added

* a `feature_powerset` jobs (using [taiki-e/cargo-hack](https://github.com/taiki-e/cargo-hack)) in the `build.yml` CI to test the features against each others

## Fixed

* the warnings that popped on my IDE when I opened the project (`cargo clippy`), using `#[allow(..)]` when meaningful.

## Changed

* use `#[cfg(all(x, y))]` for consistency in `lib.rs`

### DIFF
--- a/src/packages/string_basic.rs
+++ b/src/packages/string_basic.rs
@@ -140,13 +140,15 @@ mod print_debug_functions {
     }
 
     /// Return the empty string.
+    #[allow(unused_variables)]
     #[rhai_fn(name = "print", name = "to_string")]
-    pub fn print_unit(ctx: NativeCallContext, _unit: ()) -> ImmutableString {
+    pub fn print_unit(ctx: NativeCallContext, unit: ()) -> ImmutableString {
         ctx.engine().const_empty_string()
     }
     /// Convert the unit into a string in debug format.
+    #[allow(unused_variables)]
     #[rhai_fn(name = "debug", name = "to_debug")]
-    pub fn debug_unit(_unit: ()) -> ImmutableString {
+    pub fn debug_unit(unit: ()) -> ImmutableString {
         "()".into()
     }
 
--- a/src/packages/string_basic.rs
+++ b/src/packages/string_basic.rs
@@ -141,14 +141,12 @@ mod print_debug_functions {
 
     /// Return the empty string.
     #[rhai_fn(name = "print", name = "to_string")]
-    pub fn print_unit(ctx: NativeCallContext, unit: ()) -> ImmutableString {
-        let _ = unit;
+    pub fn print_unit(ctx: NativeCallContext, _unit: ()) -> ImmutableString {
         ctx.engine().const_empty_string()
     }
     /// Convert the unit into a string in debug format.
     #[rhai_fn(name = "debug", name = "to_debug")]
-    pub fn debug_unit(unit: ()) -> ImmutableString {
-        let _ = unit;
+    pub fn debug_unit(_unit: ()) -> ImmutableString {
         "()".into()
     }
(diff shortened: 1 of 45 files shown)


schungx:
Thanks for the very thorough job!

However, I plan to release a new version today, so I'd hesitate to make such large changes to the codebase right now.  Perhaps for the next version.

In that case, I'd suggest you split this PR into separate parts to be better manageable.  For example, one PR can add `cargo-hack` (I'm not sure how much load it'll put on the CI server).  Another PR can do tests.  Another can add clippy etc.

This way, we can apply each portion as they become ready.

Some ad hoc comments:

- I generally prefer multiple `cfg` lines instead of a single `cfg(all(...))` but this may just be personal preference.  It lets me add/remove/comment-out feature gates very easily instead of having to edit the long line.
- Certain clippy lints (e.g. complex types) I have deliberately left unfixed because I may introduce a type alias in the future...  and they sort of keep as a reminder...

Mathieu-Lala:
So, I update this PR with the latest change on `main` and removed the `cfg(all(...))`.

Let me know if some changes do not make sense :+1:

src/packages/string_basic.rs
```diff
@@ -141,14 +141,12 @@ mod print_debug_functions {
 
     /// Return the empty string.
     #[rhai_fn(name = "print", name = "to_string")]
-    pub fn print_unit(ctx: NativeCallContext, unit: ()) -> ImmutableString {
```
schungx, line 144: Better not rename parameter names because they'd go into documentation, so the documentation will have names such as `_unit` instead of `unit`.

schungx (commented):
Just that I would leave the parameter names unchanged instead of putting underscores in front of them.

src/packages/string_more.rs
```diff
@@ -84,8 +84,7 @@ mod string_functions {
     }
 
     #[rhai_fn(name = "+")]
-    pub fn add_append_unit(string: ImmutableString, item: ()) -> ImmutableString {
-        let _ = item;
+    pub fn add_append_unit(string: ImmutableString, _item: ()) -> ImmutableString {
```
schungx, line 87: Again, let's not rename parameters with underscores but put `#[allow(unused_variables)]` like in `string_basic.rs`...

schungx:
Missed a couple of underscored parameter names...

schungx:
That single failed test happens occasionally, only in CI, and only ever in that particular test.  I haven't been able to track it down yet, as it is very rarely reproducible.

If you rerun the test, it'd most likely pass.

I have a feeling it has to do with has calculations, but why it only fails on that one test remains a mystery.

schungx:
If you'd mark this PR as ready, I'll merge it.

Mathieu-Lala:
Is the failing test `test_get_set_chain_with_write_back` ?

schungx:
> Is the failing test `test_get_set_chain_with_write_back` ?

Yup.  That test just occasionally fails, and not reproducible.  I've changed some calculations in the hashing so hopefully that's what caused it.

PR #691, opened 2023-01-14
CI: add `cargo hack check --feature-powerset`

Mathieu-Lala:
# Context

Following #688 

This PR add checks concerning the cargo feature flags.

The command `cargo check --no-default-features --features bin-features,no_std` produce an build error.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,6 +88,18 @@ jobs:
         command: test
         args: ${{matrix.flags}}
 
+  feature_powerset:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@v1
+        with:
+          toolchain: stable
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-hack@0.5.25
+      - run: cargo hack check --feature-powerset --depth 2 --no-dev-deps --exclude-features "stdweb wasm-bindgen f32_float"
+
   # no-std builds are a bit more extensive to test
   no_std_build:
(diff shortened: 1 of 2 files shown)


schungx:
I think you need to exclude `no_std` as well.  It is very specialized and there is a dedicated test (sort of) for `no-std`.

schungx:
BTW, using this, will it consume a lot of CPU every time this is run?  There are a lot of features...

Mathieu-Lala:
It will definitely, this job took [17 minutes to complete with an error](https://github.com/rhaiscript/rhai/actions/runs/3918629215/jobs/6699128046). I set flag `--depth 2` as suggested on this issue https://github.com/taiki-e/cargo-hack/issues/112

schungx:
In that case, maybe we should make this manual invoke? Instead of running during every commit...

Mathieu-Lala:
Do you think this job should be run with a [`workflow_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) trigger ?

schungx:
> Do you think this job should be run with a [`workflow_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) trigger ?

That would  be a great idea!

Also, I suggest the exclude set should be:

`no_std stdweb wasm-bindgen f32_float only_i32 unicode-xid-ident bin-features`

These are the features that are sort-of "add-on"...

Mathieu-Lala:
It works like a charms https://github.com/Mathieu-Lala/rhai/actions/runs/4096543658/jobs/7064279160

PR #704, opened 2023-02-10
Doc: enhance required feature highlight on the api items

Mathieu-Lala:
Before :

image

After :

image

# Summary of changes

I made those changed only in `lib.rs` but I suppose it should be done on every public item with a `#[cfg(feature ...)]`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -147,6 +147,7 @@ instant = { version = "0.1.10" } # WASM implementation of std::time::Instant
 
 [package.metadata.docs.rs]
 features = ["document-features", "metadata", "serde", "internals", "decimal", "debugging"]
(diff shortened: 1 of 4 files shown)


schungx:
I vaguely remember that `rustdoc` will eventually put these feature flag tags onto APIs automatically...  And so I have been waiting for the feature to land.

I think if we do it, we need to add it to all the public API as you mention, which actually is not a big deal.  A lot of work, but just boilerplate type of work.  When the auto feature lands, we can just do a global search and replace it with an empty line.

Mathieu-Lala:
See
https://github.com/rust-lang/cargo/issues/8103
https://github.com/rust-lang/rust/issues/96318

PR #717, opened 2023-05-02
Fix compilation for thumbv6m-none-eabi

LinuxHeki:
I was trying to get rhai running on Raspberry Pi Pico but no-std-compat failed compiling. no-std-compat already fixed that years ago but its not yet been (and may never will be) to crates.io. First commit fixex that, second one fixes error ".swap() not found for AtomicUsize" on thumbv6m-none-eabi.

This merge request fixes compilation but I can't test it because rhai binary is larger than 2MiB (rpi pico flash size).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ bitflags = { version = "1", default-features = false }
 smartstring = { version = "1", default-features = false }
 rhai_codegen = { version = "1.5.0", path = "codegen", default-features = false }
 
-no-std-compat = { version = "0.4", default-features = false, features = ["alloc"], optional = true }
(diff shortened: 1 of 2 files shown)


schungx:
Thanks for this. Ild have to trust you on this as I dont have such a device to test on.

Also, you should be able to get a sub-MB build of rhai if you prune down features. Build for size and omit unnecessary language features.

LinuxHeki:
I got it down to 550 KiB with `Engine::RAW` and it works. `Engine::new()` doesn't work, one possible reason is not enough RAM (only 264 KiB). Raw engine is good enough for me :+1:

PR #725, opened 2023-06-14
change HokmaLock::write() to use atomic compare_exchange

garypen:
I think this fix solves [the issue we have observed on the router](https://github.com/apollographql/router/issues/3213) with hanging `Rhai` tests.

My suggested change is minimal since I want to focus on the main motivation for the change (use compare_exchange() to maintain atomic comparison), but I think this could be improved even further by more reworking. 

I realise that it's a fairly risky change, due to its importance in the code base, so I'm happy to perform any additional testing you may suggest. Our CI environment has tested it on multiple platforms.

### DIFF
--- a/src/config/hashing.rs
+++ b/src/config/hashing.rs
@@ -50,14 +50,17 @@ impl HokmaLock {
 
     pub fn write(&'static self) -> WhenTheHokmaSuppression {
         loop {
-            let previous = self.lock.load(Ordering::SeqCst);
-            self.lock.store(1, Ordering::SeqCst);
-
-            if previous != 1 {
-                return WhenTheHokmaSuppression {
-                    hokma: self,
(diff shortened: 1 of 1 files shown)


schungx:
Due to the magnitude of this bug, I'd merge it right away.

schungx:
I'll appreciate that you run more tests with this. If nothing is amiss, I'll release a new version.

garypen:
I'll keep running the tests today. At this point, we've run our router test cases many hundreds of times, but we don't use `no-std`, so I can't help much there.

Geal:
the atomic are [defined in core](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicUsize.html#method.compare_exchange) so it will have the same behaviour in std and no-std

PR #767, opened 2023-10-23
Fix repository URL for published package

bheylin:
The package metadata's repository URL refers to the Rhai organisation. It should refer to the repository that the Rhai package is publish from.


schungx:
Well, not sure...  I believe I've seen people point to the org for the main repo project...

schungx:
Ah, OK.  Fair enough.

bheylin:
Thanks!

PR #772, opened 2023-10-30
Fix VarDefInfo visibility breakage.

### LINKED ISSUE #771: v1.15 -> v1.16 breaking change: VarDefInfo field visibility
The 1.15 -> 1.16 minor version bump broke the field visibility of the `VarDefInfo` struct. This should only happen on a major version bump (1.x -> 2.x) and should ideally include deprecation warnings and a migration path in the later 1.x minor versions.

For example, both the `pub name` field and the `pub fn name()` method should be available:

```rust
pub struct VarDefInfo<'a> {
    /// Name of the variable to be defined.
    #[deprecated(since = "1.16.0", note = "VarDefInfo fields will be private in the next major release. Use the methods instead.")]
    pub name: &'a str,
...
}

impl<'a> VarDefInfo<'a> {
    /// Name of the variable to be defined.
    #[inline(always)]
    #[must_use]
    pub const fn name(&self) -> &str {
        self.name
    }
...
}
```

jrobsonchase:
Resolves #771

### DIFF
--- a/src/types/var_def.rs
+++ b/src/types/var_def.rs
@@ -7,13 +7,29 @@ use std::prelude::v1::*;
 #[derive(Debug, Clone, Hash)]
 pub struct VarDefInfo<'a> {
     /// Name of the variable to be declared.
-    name: &'a str,
+    #[deprecated(
+        since = "1.16.0",
+        note = "VarDefInfo fields will be private in the next major version. Use the methods instead."
+    )]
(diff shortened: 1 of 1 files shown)


schungx:
Thanks for catching this.  I'll put out `1.16.3` with this fix.  I won't merge this into the main branch, but I'll merge this up to `1.16.2`.

PR #790, opened 2023-12-02
fuzz: Add fuzzer to CI

silvergasp:
This change will run the fuzzer's for a short period (600s, although I'm happy to change this), on each PR.

### DIFF
--- a/.github/workflows/oss-fuzz.yml
+++ b/.github/workflows/oss-fuzz.yml
@@ -1,0 +1,42 @@
+name: CIFuzz
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


schungx:
Well, sometimes I do want them merged quickly and not wait for the results of fuzzing...  I suppose I can just press Merge even when the tests/fuzzing is not complete...

silvergasp:
Thoughts on this one. Worth merging or shall I close this one out?

schungx:
Merge it is!  Why not.  Fuzzing seems really useful!

PR #798, opened 2023-12-27
Add seed corpus entry

silvergasp:
This change should improve fuzing coverage by providing a good starting point for mutation. i.e. a script that covers the majority of the api surface.


silvergasp:
The build failures don't seem to be related to this change.

silvergasp:
This should help fill gaps in the code coverage like this one https://storage.googleapis.com/oss-fuzz-coverage/rhai/reports/20231227/linux/src/rhai/src/packages/string_more.rs.html#L171

schungx:
Not that I understand any of this, but merge away!

PR #804, opened 2024-01-01
Add dictionary for fuzz harnesses

silvergasp:


silvergasp:
For context;
- I have a corresponding change to the oss-fuzz config https://github.com/google/oss-fuzz/pull/11427
- Docs for [libfuzzer dictionaries](https://www.llvm.org/docs/LibFuzzer.html#dictionaries)

PR #815, opened 2024-01-16
Fix compile error with no 'metadata'  feature

h7kanna:
Fixed compile error as 'set_custom_type_with_comments' is gated with 'metadata' feature.

### DIFF
--- a/codegen/src/rhai_module.rs
+++ b/codegen/src/rhai_module.rs
@@ -81,27 +81,29 @@ pub fn generate_body(
             .map(syn::Attribute::to_token_stream)
             .collect();
 
+        #[cfg(not(feature = "metadata"))]
+        set_const_statements.push(
+            syn::parse2::<syn::Stmt>(quote! {
+                #(#cfg_attrs)*
+                m.set_custom_type::<#typ>(#const_literal);
+            })
+            .unwrap(),
+        );
+
         #[cfg(feature = "metadata")]
         let comments = comments
             .iter()
             .map(|s| syn::LitStr::new(s, Span::call_site()))
             .collect::<Vec<_>>();
-        #[cfg(not(feature = "metadata"))]
-        let comments = Vec::<syn::LitStr>::new();
 
-        set_const_statements.push(if comments.is_empty() {
-            syn::parse2::<syn::Stmt>(quote! {
-                #(#cfg_attrs)*
-                m.set_custom_type::<#typ>(#const_literal);
-            })
-            .unwrap()
-        } else {
+        #[cfg(feature = "metadata")]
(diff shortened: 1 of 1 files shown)


schungx:
If `metadata` is not used, then `comments` is an empty array, so the branch with `set_custom_type_with_comments` is never called.  I don't particularly see a bug here...

h7kanna:
Below issue with

```
rhai = { version = "1.16.3", features = ["internals", "no_closure", "no_module", "no_custom_syntax"], optional = true }
```

Screenshot 2024-01-16 at 8 50 04 PM

schungx:
What version of `rhai_codegen` are you using?  Can you do a `cargo update`?

h7kanna:
I am not using rhai_codegen directly. Now I realize that the usage of 'rhai' alone compiles just fine.
May be, This is just IntelliJ IDE messing up with features.

schungx:
> I am not using rhai_codegen directly. Now I realize that the usage of 'rhai' alone compiles just fine. May be, This is just IntelliJ IDE messing up with features.

`rhai` will call `rhai_codegen` which is why you need to be careful about version conflicts.  I suspect you have a lower version of `rhai_codegen` in your `Cargo.lock`...

Also, if the IDE is working on an old version of `rhai_codegen` that may also cause this problem...

h7kanna:
But still, for the code to work [set_fn_with_comments](https://github.com/rhaiscript/rhai/blob/main/src/module/mod.rs#L1181)  will not be in scope without 'metadata' feature. no?

schungx:
> But still, for the code to work [set_fn_with_comments](https://github.com/rhaiscript/rhai/blob/main/src/module/mod.rs#L1181) will not be in scope without 'metadata' feature. no?

Not, the codegen won't generate this function call without the `metadata` feature.

h7kanna:
Cargo.lock

```
[[package]]
name = "rhai_codegen"
version = "1.6.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "853977598f084a492323fe2f7896b4100a86284ee8473612de60021ea341310f"
dependencies = [
 "proc-macro2",
 "quote",
 "syn 2.0.48",
]
```

h7kanna:
> > But still, for the code to work [set_fn_with_comments](https://github.com/rhaiscript/rhai/blob/main/src/module/mod.rs#L1181) will not be in scope without 'metadata' feature. no?
> 
> Not, the codegen won't generate this function call without the `metadata` feature.

Then its IDE not setting features correctly

h7kanna:
This is just IntelliJ dealing badly with feature unification. Closing this.

PR #817, opened 2024-01-20
Add a derive macro for CustomType

MavethGH:
The macro currently generates getters and setters for all fields in the struct, and has helper attributes for overriding generated getter/setter functions. There is also a `readonly` helper attribute that prevents a setter from being registered.

TODOs:
- [ ] Docs docs docs
- [ ] More tests
- [ ] Register a pretty-print name for types
- [ ] Attribute for calling a user-provided function with the `TypeBuilder` passed in
    - What would it be called?
    - Do we call it before or after registering everything else?
- [ ] Attribute for skipping a field entirely
- [ ] Support for tuple structs
    - What would the fields be named? `field0`, `field1`, etc perhaps?
- [ ] Support for enums as [described in the book](https://rhai.rs/book/patterns/enums.html)

### DIFF
--- a/codegen/src/custom_type.rs
+++ b/codegen/src/custom_type.rs
@@ -48,8 +48,8 @@ pub fn derive_custom_type_impl(input: DeriveInput) -> TokenStream {
 
     quote! {
         impl ::rhai::CustomType for #name {
-            fn build(builder: ::rhai::TypeBuilder<'_, Self>) {
-                #accessors
+            fn build(mut builder: ::rhai::TypeBuilder<'_, Self>) {
+                #accessors;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


schungx:
Looking good!

Let me know when to merge.

schungx:
So below is list of all the attributes so far...

* All fields (`pub` or private) will be exported
    * perhaps later a `skip` attribute
* `#[readonly]` => only getter, no setter
* `#[get]` => generated getter
    * `#[get(func_name)]` => use `func_name` as getter function
* `#[set]` => generated setter
    * `#[set(func_name)]` => use `func_name` as setter function
* omitted => generated getter and setter

schungx:
Unfortunately we cannot make `#[get(func_name)]` use a standard getter signature with `&self` instead of `&mut self` which is quite unusual for normal structs...

schungx:
I'll open another tracking issue...

PR #829, opened 2024-01-31
Added alias to ThinVec with internals feature

h7kanna:
Added alias to avoid importing [thinvec](https://docs.rs/thin-vec/latest/thin_vec/) separately as `Expr::InterpolatedString` accepts ThinVec now.

### DIFF
--- a/src/ast/ast.rs
+++ b/src/ast/ast.rs
@@ -1,7 +1,7 @@
 //! Module defining the AST (abstract syntax tree).
 
 use super::{ASTFlags, Expr, FnAccess, Stmt};
-use crate::{Dynamic, FnNamespace, ImmutableString, Position};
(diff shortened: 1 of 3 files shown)


schungx:
It would be better to take after `StaticVec` where the type is re-exported no matter what, only with `pub` if `internals` is on.

Otherwise your code would fail if `internals` is off.  Do a `cargo check` to confirm.

This way, we can standardize on `crate::ThinVec` through-out the code.  There are quite a few `thin_vec::ThinVec` in there that you should scan through, and a few more places of `use thin_vec::ThinVec`.

schungx:
It is probably easier that I do it, if you prefer.

h7kanna:
> It is probably easier that I do it, if you prefer.

Sure.

schungx:
> > It is probably easier that I do it, if you prefer.
> 
> Sure.

Alright, I'll close this.

PR #840, opened 2024-02-26
Parse json within a script

Mathieu-Lala:


schungx:
Thanks for the PR!

Mathieu-Lala:
I managed to make the CI pass, with one basic test of the `parse_json` function and two error case: 
* a json array under `no_index`
* a call to `parse_json` under `no_object`

Let me know if something is missing to this PR :+1:

schungx:
How about using `serde_json` under `metadata`?  Since metadata pulls it in, it can use `serde_json` directly...

So gate that also under `#[cfg(not(feature = "metadata"))]`, and put in another implementation using `serde_json` under `metadata`?

schungx:
Looking at it, the two can probably be merged into a single function:
```rust
    #[cfg(not(feature = "no_object"))]
    #[rhai_fn(return_raw)]
    pub fn parse_json(ctx: NativeCallContext, json: &str) -> RhaiResultOf<crate::Map> {
        #[cfg(feature = "metadata")]
        return serde_json::from_str(json).map_err(|err| err.to_string().into());
        #[cfg(not(feature = "metadata"))]
        return ctx.engine().parse_json(json, true);
    }
```

Mathieu-Lala:
I made a mistake, i should had rename the parmeter `_ctx` for `ctx` in the `parse_json`

PR #845, opened 2024-03-18
feat: range indexing on strings

### LINKED ISSUE #841: Feature: syntax `str[range]` support like rust string indexing by range
Expect:

```rust
let s = "abcd";
let n = s[1..3];
```

Now it reports:

```text
2: let n = s[1..3];
             ^ Data type incorrect: core::ops::range::Range<i64> (expecting i64)
```

Would this feature could be supported? Maybe I can contribute if you don't mind to add this, and give me some help to start.

zitsen:
Now we support range indexing on strings, including ref/mut impls.

For example, you can use range indexing intead of sub_of_string() method like bellow:

```rust
let s1 = "abcde";
let s2 = s1[1..4]; // "bcd"
s1[1..4] = "f"; // s1: "afe", s2: "bcd"
```

Close #841

### DIFF
--- a/src/ast/expr.rs
+++ b/src/ast/expr.rs
@@ -530,6 +530,9 @@ impl Expr {
                             Self::IntegerConstant(ref start, ..),
                             Self::IntegerConstant(ref end, ..),
                         ) => (*start..*end).into(),
+                        (Self::IntegerConstant(ref start, ..), Self::Unit(..)) => {
+                            (*start..INT::MAX).into()
+                        }
                         _ => return None,
                     },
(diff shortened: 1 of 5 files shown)


schungx:
Wow, I'll look through that.

Thanks!

zitsen:
I've test it with `cargo test` to confirm it wouldn't break anything. For char indexing, I've add a more complex test case to test both char-index and range-index on the same string.

schungx:
Anyway let's merge this and see how it goes.

PR #847, opened 2024-03-20
Documentation generation for custom types

ltabis:
Add documentation for types that derive the `CustomType` macro with the `metadata` feature.


schungx:
Still too many `_with_comments` for my comfort...  I went down this route before and it took a lot of breaking changes and deprecation to finally move them all into the `FuncRegistration` API... I would hesitate to add a whole bunch of `_with_comments` yet again.

I am wondering if there is a way to do this without introducing a parallel set of API's... 

How about if we include "the last function hash" inside `TypeBuilder`.  Then you can do:

```rust
builder
    .with_get("foo", ...)      // stores hash in builder type
    .add_comments(...)     // always uses hash in builder (if Some) to look up function in engine
    ....
```

ltabis:
@schungx I think this is a good idea ... for functions. I guess that for types we still need a `with_name_and_comments` function since the engine does not store hashes for them.

ltabis:
I fiddled a bit with the FunctionRegistration API to see if I can keep an instance of `FuncRegistration` in the builder, unfortunately with the current builder implementation there is not `build` function that tells the builder to register the current function.

However there is this deprecated `Module::update_fn_metadata_with_comments` function that I could use as a `add_fn_comments` function like you said. Could I just remove it from deprecated an just add a `pub(crate)` visibility on it so that only the builder API can use it ?

schungx:
Since each builder has only one type, it is quite trivial to add a `with_comments` for the builder.

Implementation wise, it can lookup the registered type via `TypeId` and add comments to it.

This way no public API needs to change. Only a few more methods to the builder API.

ltabis:
I have done the changes, however when I execute the `custom_types_and_methods` example the function parameters and return type information seems lost. I'm looking for the cause of the problem.

ltabis:
also rustc 1.77 seems to break the `ui_tests/rhai_mod_inner_cfg_false.rs` test

ltabis:
> I have done the changes, however when I execute the `custom_types_and_methods` example the function parameters and return type information seems lost. I'm looking for the cause of the problem.

Find out why. All the documentation steps made in `Engine::register_fn` are not done when using the `FuncRegistration` API directly. I used the `FuncRegistration` API because I can get the hash of the function directly when it's registered in the engine. I cannot get it using `register_fn`.

ltabis:
> I have done the changes, however when I execute the `custom_types_and_methods` example the function parameters and return type information seems lost. I'm looking for the cause of the problem.

Find out why: using the `FuncRegistration` API directly does not set the `params_info` by itself.

schungx:
Looking good!

A couple of observations:

* Since an empty `Vec` indicates no hashes, you can omit the `Option`

* Probably not a good idea to move the default params info into `register_into_engine`... What if the user used `with_params_info` before? Those would be overwritten...

ltabis:
> A couple of observations

I agree with your comments, I made the changes.

schungx:
I think it looks good. Let's merge this.

schungx:
I wonder if it is possible to avoid adding `update_fn_comments`...

Since the builder API is in the same crate, I think I can just inline the code in.

PR #866, opened 2024-04-30
Make `Dynamic::try_cast_raw` public

paholg:
This function is useful for error-reporting. With `try_cast`, one has to do one of the following:
1. Pre-clone the `Dynamic`.
2. Check `Dynamic::is` first, and still handle the `None` case of `try_cast`.

### DIFF
--- a/src/api/call_fn.rs
+++ b/src/api/call_fn.rs
@@ -183,7 +183,7 @@ impl Engine {
             options,
         )
         .and_then(|result| {
-            result.try_cast_raw().map_err(|r| {
+            result.try_cast_result().map_err(|r| {
                 let result_type = self.map_type_name(r.type_name());
                 let cast_type = match type_name::<T>() {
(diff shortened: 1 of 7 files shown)


schungx:
You have a better name for `try_cast_raw`?  I tend to try to keep `_raw` for internal API or some really advanced API...

paholg:
Fair enough. If I could go back in time, I would propose this function be `try_cast` and the one that returns an option be renamed or removed, but that is not worth a breaking change.

I can't think of a great name. Maybe `try_cast_res` or `try_cast_result`?

schungx:
Picking a good name is important for an API, just for that reason you just mentioned -- we can never ever change it.

Let's spend a couple more days thinking up new names...  there is no rush until the next release anyway...

paholg:
I have updated the name to `try_cast_result`.

schungx:
Alright let's merge this.

PR #874, opened 2024-05-14
Fix #867

### LINKED ISSUE #867: The `sync` feature flag is not compatible with `no_std`
when i enable both the `sync` and the `no_std` feature flags, rhai won't compile due to using `std::sync::RwLock`, which is not provided by the `no-std-compat` crate due to not using the `compat_sync` feature flag of that crate.

this can simply be solved by enabling the `compat_sync` feature flag of the `no-std-compat` crate.

misssonder:
try to fix https://github.com/rhaiscript/rhai/issues/867

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ bitflags = { version = "2.0.0", default-features = false }
 smartstring = { version = "1.0.0", default-features = false }
 rhai_codegen = { version = "2.1.0", path = "codegen" }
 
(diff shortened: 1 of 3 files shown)


schungx:
Not sure about the code changes other than turning on the feature flag, do I'm running the tests.

schungx:
Oops, shouldn't merge so quickly.  There is a failed test that I didn't catch.

The `sync` feature doesn't build.  It requires `unwrap`.

Issue #899, opened 2024-07-15
Unhelpful error when trying to call a non-existing function or with wrong signature


therealprof:
The returned error when trying to call a function implemented in Rust with either incorrect name or incorrect signature is not helpful. E.g.

```Function not found: add_param_setting (rusp::usp_builder::add::CreateObjectBuilder, &str | ImmutableString | String, &str | ImmutableString | String, i64) (line 5, position 10)```

It's basically throwing back at you what you're trying to do without any hints about what might be wrong with it.

I guess in the interest of space we don't want to do some guesswork to find the correct fix and make a suggestion. But it would be nice to at least detect whether the function doesn't exist at all (i.e. wrong function name) or the function does exist but has a non-matching signature.

schungx:
Well, it does print out the parameter types so you can find the function that is missing...

What other info do you suggest?

therealprof:
> Well, it does print out the parameter types so you can find the function that is missing...

It takes quite a bit of work to look into the Rust code to figure out what's wrong. People who want to write Rhai scripts don't necessarily know how to navigate the the Rust implementation to figure out the source of an error. There're some additional pitfalls, things like numbers always being as `i64` which trip even me frequently.

I was thinking of changing all arguments into `Dynamic` and returning `Result`s so I can create more useful error messages, but that seems very heavy handed...

> What other info do you suggest?

At the very least I would split the one error message into two:
- Unknown function
- Call arguments don't match signature

Ideally the signature mismatch would result in a message pointing out the signature of the function that was found with that name, but I don't know whether that's easily possible or too complex.

schungx:
Well the point is... Function not found is the same as wrong parameter types. Rhai functions can be overloaded.

therealprof:
For me adding suggestions would be a nice bonus but the real important bit is: As a user I want to know whether a function of that name doesn't exist at all or there is at least one such function but the combination of passed-in arguments in didn't match any of the function signatures.

If you write a shell script and you get a `command not found` error, you'll check for a completely different issue than if you had gotten an `invalid arguments` error.

The same is true for Rhai, especially with complex signatures. It took me quite some time to figure out that my script stopped working in the middle of nowhere just because of a typo in the function name while I was meticulously checking the correctness of the passed in arguments against the signature of the function I would have wanted to call, if there hadn't been the typo.

[...]

PR #900, opened 2024-07-16
Silence some compiler warnings about unused imports

therealprof:
While trying something with the tests I noticed that there're some unused import warnings which can easily be silenced.

### DIFF
--- a/tests/arrays.rs
+++ b/tests/arrays.rs
@@ -1,6 +1,6 @@
 #![cfg(not(feature = "no_index"))]
-use rhai::{Array, Dynamic, Engine, EvalAltResult, ParseErrorType, Position, INT};
-use std::{convert::TryInto, iter::FromIterator};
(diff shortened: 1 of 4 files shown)


schungx:
Well, currently the coding style is that unused imports are OK in tests and other non-lib places.

Rhai has a large number of features and silencing all warnings will need to explicitly qualify a lot of things in lots of places.

We try to silence warnings for lib code so projects using Rhai won't get unnecessary compile warnings

therealprof:
I do find the warnings very distracting as I was looking into improving the #899 situation and I'm also not attempting to fix the remaining 2 or 3 gnarlier ones. But I'm fine closing this PR if you don't want to address this cosmetic issue.

schungx:
Let's run the CI and see...

schungx:
Closing this for now.

PR #901, opened 2024-07-16
Rhai shared scoped refs

mrazorvin:


schungx:
A couple of quick observations:

- Try not to change existing API like `register_get` or else we'll break code. Better to add to the API surface instead.

- Same for examples and tests... Try to add instead of change. This would be quite messy to undo by the time we're ready to merge.

- Try to have as few changed files as possible as that would help with reasoning. Also it'll help putting this under a feature flag.

schungx:
I wonder how this is working out...

ImTheSquid:
Stale

PR #902, opened 2024-07-23
Add ref_* methods for Dynamic.

madonuko:
These methods does not take ownership and are easier to work with than their `into_*` counterparts.

### DIFF
--- a/src/types/dynamic.rs
+++ b/src/types/dynamic.rs
@@ -2535,7 +2535,7 @@ impl Dynamic {
     ///
     /// These normally shouldn't occur since most operations in Rhai are single-threaded.
     #[inline]
-    pub fn get_string(&self) -> Option<DynamicReadLock<String>> {
+    pub fn get_immutable_string(&self) -> Option<DynamicReadLock<ImmutableString>> {
         self.read_lock()
     }
     /// Convert the [`Dynamic`] into an [`ImmutableString`].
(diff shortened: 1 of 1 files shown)


schungx:
Can these be replaced by `lock_read`?

schungx:
Come to think of it and experimenting with a number of different names, I suggest `as_XXX_ref` and `as_XXX_ref_mut` in order to avoid conflicts with a deprecated API called `as_immutable_string`.

I have actually put together a PR here:  https://github.com/rhaiscript/rhai/pull/904

Would you wait until I commit this, then rebase your further changes on top of it?

madonuko:
Sure, I'm happy to wait. Also, why not `as_XXX_mut` (following rust std)?

schungx:
> Sure, I'm happy to wait. Also, why not `as_XXX_mut` (following rust std)?

Hhhmmm... Good idea. I'll change.

schungx:
Changed in https://github.com/rhaiscript/rhai/pull/905

schungx:
Is it necessary to keep the `get_XXX` versions since the only difference now is `Option` vs `Result`...

madonuko:
Well I guess the compiler can optimize the error handling away while compiling this crate?

schungx:
> Well I guess the compiler can optimize the error handling away while compiling this crate?

Yes it can...  but I'm not sure having `get_string()` would be significant better than `as_immutable_string_ref().ok()`...  it is quite a bit shorter, true, but also very similar in API...

For others, `get_array()` and `as_array_ref().ok()` are not as pronounced in the length of the call...

schungx:
I would close this for the time being until later date when more API needs to be added.

Issue #891, opened 2024-07-03
Rhai depends on unmaintained crate "instant"


VorpalBlade:
According to https://crates.io/crates/instant:

> This crate is no longer maintained. Please consider creating a fork or using `web-time` instead. Or reach out if you are interested in taking over its maintenance.

rhai should probably switch?

schungx:
Hhhhmmm.... good catch.

Not sure if `web-time` would be a drop-in replacement...

As `instant` is primarily used only in WASM and I don't usually build for WASM...  Somebody needs to test it out and let me know if everything is fine if we replace `instant` with `web-time`...

VorpalBlade:
> As `instant` is primarily used only in WASM and I don't usually build for WASM... Somebody needs to test it out and let me know if everything is fine if we replace `instant` with `web-time`...

Same, I was just adding rhai to a new project and was auditing the recursive dependencies it pulled in and noticed this.

misssonder:
> As `instant` is primarily used only in WASM and I don't usually build for WASM... Somebody needs to test it out and let me know if everything is fine if we replace `instant` with `web-time`...

If  use `web-time` instead of `instant`, will it cause a breaking change?

schungx:
> > As `instant` is primarily used only in WASM and I don't usually build for WASM... Somebody needs to test it out and let me know if everything is fine if we replace `instant` with `web-time`...
> 
> If use `web-time` instead of `instant`, will it cause a breaking change?

Not sure... I haven't used `web-time` before... can somebody test it out?

TrueDoctor:
Could we make pulling in time optional? We are currently thinking about adding Rhai support to graphite: https://github.com/GraphiteEditor/Graphite/pull/2460 but in our case we don't need time support at all so it would be great if we could just disable this altogether

schungx:
Try `no_time`?

PR #907, opened 2024-08-05
Add web-time

misssonder:
according to https://github.com/rhaiscript/rhai/issues/891, since [instant](https://crates.io/crates/instant) is no longer maintained, we should use [web-time](https://crates.io/crates/web-time) to replace it. Actually I have built WASM to make sure it works in it, but it seems that I don't find unit tests for WASM.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -156,7 +156,8 @@ codegen-units = 1
 #panic = 'abort'     # remove stack backtrace for no-std
 
 [target.'cfg(target_family = "wasm")'.dependencies]
-web-time = { version = "1.1.0" }# WASM implementation of std::time::Instant
+[target.'cfg(not(feature = "no_time"))'.dependencies]
+web-time = { version = "1.1.0" ,optional = true}# WASM implementation of std::time::Instant
 
 [package.metadata.docs.rs]
 features = ["document-features", "metadata", "serde", "internals", "decimal", "debugging"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -118,9 +118,9 @@ no_std = ["no-std-compat", "num-traits/libm", "core-error", "libm", "hashbrown",
 #! ### JavaScript Interface for WASM
 
 ## Use [`wasm-bindgen`](https://crates.io/crates/wasm-bindgen) as JavaScript interface.
-wasm-bindgen = ["getrandom/js", "instant/wasm-bindgen"]
+wasm-bindgen = ["getrandom/js"]
 ## Use [`stdweb`](https://crates.io/crates/stdweb) as JavaScript interface.
-stdweb = ["getrandom/js", "instant/stdweb"]
+stdweb = ["getrandom/js"]
 
 #! ### Features used in testing environments only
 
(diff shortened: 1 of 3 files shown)


schungx:
It is failing the `no-std` build... Can you check?

Also I wonder if it is possible to write tests for WASM...

misssonder:
> Also I wonder if it is possible to write tests for WASM...

I'm not sure if I can write tests for WASM, but I'll try it. BTW, I have fixed the `no-std` build.

misssonder:
If we add `web-time` with optional = `true`  then we can't pass the `cargo build --target wasm32-unknown-unknown --no-default-features` compile; Otherwise if we set optional = `false`, we can't pass `cargo build --manifest-path=no_std/no_std_test/Cargo.toml --profile unix ` compile. It seems like there isn't perfect way to solve it.

schungx:
Why is the wasm build failing without default features?

schungx:
`[target.'cfg(not(feature = "no_time"))'.dependencies]`

Why did you change it to no_time instead? Before it is pulled in only for wasm which doesnt have `Instant`.

misssonder:
> `[target.'cfg(not(feature = "no_time"))'.dependencies]`
> 
> Why did you change it to no_time instead? Before it is pulled in only for wasm which doesnt have `Instant`.

You are correct. I have changed it to `target.'cfg(all(target_family = "wasm", not(feature = "no_time")))'.dependencies`, but it still fails the tests. It appears that when we add `web-time` as dependencies, it fails the no_std tests. I'm not certain if it's because of the implementation of `web-time`.

schungx:
It shoudnt... no-std is not supposed to pull in web-time, then it shouldnt be affected.

If it still fails that means web-time is still being brought in somehow...

misssonder:
I'm pretty sure the issue is related to `once_cell` and `web_time`. I've created an [example](https://github.com/misssonder/oncecell-with-webtime-example) to reproduce the problem. Can run `cargo c --features time` or `cargo c` to see the issue. However, it works correctly if I switch from `web_time` to `instant`. I'm not sure why this is happening.

schungx:
Is it possible that `web-time` does not support `no-std`?

misssonder:
It's not supported to be pulled if I don't add features `time` in [this](https://github.com/misssonder/oncecell-with-webtime-example/blob/main/Cargo.toml#L13), but it still failed.

schungx:
> It's not supported to be pulled if I don't add features `time` in [this](https://github.com/misssonder/oncecell-with-webtime-example/blob/main/Cargo.toml#L13), but it still failed.

If only one dependency crate in the entire tree depends on `std`, then `stdlib` will be pulled in, causing compilation error in `no-std`.

misssonder:
It turn out that `feature = ...` in `target.'cfg(...)'.dependencies` is not supported for selecting dependencies and will not work as expected. So actually `rhai` will always pull `instant` whenever it's `wasm` or not. But `web-time` don't support to work in `no_std` environment, so now we can't use `web_time` to replace `instant`.

lexi-the-cute:
> It turn out that `feature = ...` in `target.'cfg(...)'.dependencies` is not supported for selecting dependencies and will not work as expected. So actually `rhai` will always pull `instant` whenever it's `wasm` or not. But `web-time` don't support to work in `no_std` environment, so now we can't use `web_time` to replace `instant`.

I wonder if setting to `resolver = 2` will fix that issue. https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#cargos-new-feature-resolver

yonas:
Probably should mark this as a duplicate of https://github.com/rhaiscript/rhai/pull/1006

PR #919, opened 2024-09-30
Fix compile on thumbv6m

chxry:


chxry:
`critical-section` isn't supported on most targets, using it should be under a feature flag.

schungx:
Oops. I merged it already.

Maybe I should reverse it?

schungx:
But if it is available on a low end ARM, I suppose it is universal enough?

chxry:
It only really exists on embedded systems, that change breaks building no_std_test. I've fixed it [here](https://github.com/chxry/rhai/commit/09bccd032d1c3f151c2d51a4053372b5d5fb748f) and can create a new PR.

schungx:
Pls do. I don't do embedded and so I cannot test myself.

Issue #869, opened 2024-05-05
Rhai `no_std` appears to rely on `once_cell/alloc` but does not request it


cbiffle:
Hi! Playing with Rhai on `no_std` over here (on an actual platform that doesn't provide `std`).

Currently, 1.18.0 won't build on a platform that doesn't provide `std`, because it appears to be relying on the `once_cell` crate having the `alloc` feature set (to get `once_cell::race::OnceBox`), but isn't setting that feature itself -- at least when depended on as

```
rhai = { version = "1.18.0", default-features = false, features = ["no_std"] }
```

Adding a redundant dep to `once_cell` and setting the `alloc` feature in my build causes it to start working:

```
# TODO: had to add this because rhai isn't requesting a feature it relies on.
once_cell = { version = "1.19.0", default-features = false, features = ["alloc"] }
```

cbiffle:
From the commit history it looks like this might be a known bug; might be worth adding something to the README that the packaged versions require this workaround, lest it trip up more people. Feel free to close if you feel it's fixed.

schungx:
@cbiffle  do you have more info on this?

PR #924, opened 2024-10-23
Use Locked constructor instead of into

stargazing-dino:
I don't know why but when I was building this lib for my little micro controller in no_std, I was getting this error:

```
error[E0277]: the trait bound `RwLock<StringsInterner>: core::convert::From<StringsInterner>` is not satisfied
  --> /Users/katzen/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rhai-1.19.0/src/api/mod.rs:66:68
   |
66 |             self.interned_strings = Some(StringsInterner::new(max).into());
   |                                                                    ^^^^ the trait `core::convert::From<StringsInterner>` is not implemented for `RwLock<StringsInterner>`, which is required by `StringsInterner: core::convert::Into<_>`
   |
   = note: required for `StringsInterner` to implement `core::convert::Into<RwLock<StringsInterner>>`
```

So I tried just pulling in the Locked type and using that instead. Seemed to fix things.

Here is my a snippet of my dependency list if it's interesting:

```toml
[dependencies]
# TODO: had to add this because rhai isn't requesting a feature it relies on.
once_cell = { version = "1.19.0", default-features = false, features = [
    "alloc",
] }
# https://rhai.rs/book/start/builds/minimal.html
rhai = { version = "1.19.0", default-features = false, features = [
    "no_std",
    "sync",
    "only_i32",
    "f32_float",
    "no_custom_syntax",
    "no_time",
] }
```

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -35,7 +35,7 @@ pub mod deprecated;
 
 use crate::func::{locked_read, locked_write};
 use crate::types::StringsInterner;
-use crate::{Dynamic, Engine, Identifier};
+use crate::{Dynamic, Engine, Identifier, Locked};
 
 #[cfg(feature = "no_std")]
 use std::prelude::v1::*;
@@ -63,7 +63,7 @@ impl Engine {
                 guard.set_max(max);
             }
         } else {
-            self.interned_strings = Some(StringsInterner::new(max).into());
+            self.interned_strings = Some(Locked::new(StringsInterner::new(max)));
         }
         self
     }


schungx:
It could be that `RwLock` in `no-std` has some compatibility issues...

However we have been liberally using `.into()` everywhere so I wonder why it only trips up here...

BTW which microcontroller are you building for? We don't have many examples of `no-std` builds and I'd love if you can share some snippets.

Also the need for `once_cell/alloc` is for your own code, correct? Thats because it doesnt seem to require it in a normal `no-std` build.

stargazing-dino:
Yeah, honestly no clue why it tripped up here. Once I get home I'll try again from scratch and see if anything changes. Might've been a temporary thing. I did make sure to update cargo/rust though and my dependencies to latest.

The once_cell thing came from https://github.com/rhaiscript/rhai/issues/869 and I don't think it's actually needed. I removed it yesterday and it didn't have a hiccup but at one point it did fix a similar issue for me.

I am using an nrf52. Not really much code as it's all spaghetti nonsense right now but I could share with you a minimal example perhaps running on a microbit v2 later this week.

schungx:
> The once_cell thing came from #869 and I don't think it's actually needed. I removed it yesterday and it didn't have a hiccup but at one point it did fix a similar issue for me.

Yeah, I remember there were some changes since then.

BTW, appreciate if you can dig out why it failed to compile on that particular line, because I can't seem to reproduce it.

stargazing-dino:
Tried it again with a clean compile and it went through no problem. No idea why this was erroring at first :\

Anyways, I'll close this as I don't think it's needed

jeromegn:
I'm compiling for the `x86_64-unknown-none` target and getting a similar issue with `sync` and `no_std` enabled when using a `once_cell::sync::Lazy`:

```
error[E0277]: the trait bound `RwLock<StringsInterner>: core::convert::From<StringsInterner>` is not satisfied
  --> /home/jerome/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rhai-1.20.0/src/api/mod.rs:66:68
   |
66 |             self.interned_strings = Some(StringsInterner::new(max).into());
   |                                                                    ^^^^ the trait `core::convert::From<StringsInterner>` is not implemented for `RwLock<StringsInterner>`
   |
   = note: required for `StringsInterner` to implement `core::convert::Into<RwLock<StringsInterner>>`
```

Unfortunately, `cargo clean` did not fix it :)

Maybe minimal repo:

```rust
#![no_std]

use once_cell::sync::Lazy;

static ENGINE: Lazy<rhai::Engine> = Lazy::new(rhai::Engine::new_raw);

fn main() {}
```

```
[dependencies]
once_cell = { version = "1.20.2", default-features = false, features = ["critical-section"] }
rhai = { version = "1.20.0", default-features = false, features = ["no_std", "sync"] }

[profile.release]
panic = "abort"
```

schungx:
This is totally strange... I'm happy to just put in the fiz, but I fear we may be masking a deeper, more subtle bug.

Issue #928, opened 2024-11-02
Add `call_fn`, `call_native_fn` etc. methods to `EvalContext`


rawhuul:
It would be beneficial to enable on-the-fly function calls when using custom syntax, as previously suggested in issue #923.

PR #929, opened 2024-11-02
Added `call_fn`, `call_native_fn`, `call_fn_raw` and `call_native_fn_raw` to `EvalContext`

rawhuul:
Addresses #928

### DIFF
--- a/src/eval/eval_context.rs
+++ b/src/eval/eval_context.rs
@@ -1,7 +1,15 @@
 //! Evaluation context.
 
 use super::{Caches, GlobalRuntimeState};
-use crate::{expose_under_internals, Dynamic, Engine, Scope};
+use crate::{
+    ast::FnCallHashes,
+    expose_under_internals,
+    func::calc_fn_hash,
+    tokenizer::{is_valid_function_name, Token},
+    types::Variant,
+    Dynamic, Engine, FnArgsVec, FuncArgs, Position, RhaiResult, RhaiResultOf,
+    Scope, StaticVec, ERR,
+};
 #[cfg(feature = "no_std")]
 use std::prelude::v1::*;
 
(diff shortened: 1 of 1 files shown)


rawhuul:
> Great stuff. However, I suggest we defer these functions to the NativeCallContext methods instead of  duplicating them? Do you think it can be done? In other words, creating a NativeCallContext from a CallContext...

Got it! So basically, we need to create a streamlined API that will provide `NativeCallContext` instead of `call_fn` methods directly, right? I’ll definitely work on that.

But I have a question: if `NativeCallContext` already stores the function name, why do we still need to pass the function name as an argument when calling the `call_fn*` functions?

schungx:
> Got it! So basically, we need to create a streamlined API that will provide `NativeCallContext` instead of `call_fn` methods directly, right? I’ll definitely work on that.

Perhaps duplicate the methods, but inside the implementation create a `NativeCallContext` and delegate.  That might be simpler to the user than having them create a `NativeCallContext`...  although with proper docs it won't be a prob.

> But I have a question: if `NativeCallContext` already stores the function name, why do we still need to pass the function name as an argument when calling the `call_fn*` functions?

That's because the function name is the name of the _current_ function, while the name passed to `call_fn` is the function that the current function intends to call...

rawhuul:
> Perhaps duplicate the methods, but inside the implementation create a NativeCallContext and delegate. That might be simpler to the user than having them create a NativeCallContext... although with proper docs it won't be a prob.

Yah, cool. I'll surely look into it.

> That's because the function name is the name of the current function, while the name passed to call_fn is the function that the current function intends to call...

Okay, but it seems that the `NativeCallContext` works with an empty string, it could be `Option<&str>` then.

schungx:
There seems to be a few tricky points regarding this implementation so I'd close this and redo in a new PR.

Issue #547, opened 2022-04-01
is there any way to serialize and deserialize rhai ast from and to string?


OfekShochat:
hello! Im trying to cache asts but I don't see any way to do this. I don't think #161 is finished, but idk. is there any standard way of doing this or should I try and take all the info and then reconstruct the ast?
thanks!
ps great project

schungx:
Well, I actually tried something like this in my on fork, in a branch called `ast_serde`.  It is very old though.

https://github.com/schungx/rhai/tree/ast_serde

At half ways, I realized that there really is no point in doing this, as persisting the original script is easier and quicker than generating a string serialization from an `AST`.

Therefore, the standard way is to persist the script text itself (which can be compressed easily if you're really pressed for space).

PR #936, opened 2024-11-27
Add support for serializing/deserializing AST statements

fluxlinkage:
- Why?

[The Rhai book](https://rhai.rs/book/patterns/serialize-ast.html) has already explained that serializing an AST leads to larger storage cost and possibly slower restoring speed than recompiling the original script. Therefore, it is generally not recommended.

However, Serializing/deserializing AST may still be useful under certain conditions. For example, by distributing serialized ASTs instead of original scripts, you can prevent users from seeing the source code (especially when using a binary format), while Rhai scripts still work properly.

- Example

Cargo.toml

```toml
[package]
name = "rhai_ast_serde_test"
version = "0.1.0"
edition = "2021"

[dependencies]
serde = {version="1"}
serde_json = {version="1"}
rhai = { path = "../rhai" ,features=["ast-serde"]}
```

main.rs

```Rust
use rhai::{Engine, EvalAltResult, Stmt};

fn main() -> Result<(), Box<EvalAltResult>> {
    let script = "// Rhai script you want to run. For example:
let now = timestamp();
let x = 1_000_000;

print(\"Ready... Go!\");

while x > 0 {
    x -= 1;
}

print(`Finished. Run time = ${now.elapsed} seconds.`);";
    let engine = Engine::new();
    let ast = engine.compile(script)?;
    println!("=== Original AST ===");
    println!("{:?}", ast);
    let json = serde_json::to_string(ast.statements()).unwrap();
    println!("=== JSON ====");
    println!("{}", json);
    let stmts_restored = serde_json::from_str::<Vec<Stmt>>(&json).unwrap();
    let ast_restored = rhai::AST::new(stmts_restored, rhai::Module::default());
    println!("=== Restored AST ===");
    println!("{:?}", ast_restored);
    println!("=== Outputs ===");
    engine.run_ast(&ast_restored)?;

    Ok(())
}
```

- References

https://github.com/rhaiscript/rhai/issues/547

https://github.com/schungx/rhai/tree/ast_serde

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,6 +71,8 @@ debugging = ["internals"]
 bin-features = ["decimal", "metadata", "serde", "debugging", "rustyline"]
 ## Enable fuzzing via the [`arbitrary`](https://crates.io/crates/arbitrary) crate.
 fuzz = ["arbitrary", "rust_decimal/rust-fuzz", "serde"]
+## Enable serialization/deserialization of AST.
+ast_serde = ["serde","bitflags/serde"]
 
 #! ### System Configuration Features
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,7 +72,7 @@ bin-features = ["decimal", "metadata", "serde", "debugging", "rustyline"]
 ## Enable fuzzing via the [`arbitrary`](https://crates.io/crates/arbitrary) crate.
 fuzz = ["arbitrary", "rust_decimal/rust-fuzz", "serde"]
 ## Enable serialization/deserialization of AST.
-ast_serde = ["serde","bitflags/serde"]
+ast-serde = ["internals","serde","bitflags/serde"]
 
 #! ### System Configuration Features
 
--- a/src/ast/expr.rs
+++ b/src/ast/expr.rs
@@ -23,6 +23,7 @@ use std::{
 /// _(internals)_ A binary expression.
 /// Exported under the `internals` feature only.
 #[derive(Debug, Clone, Hash, Default)]
+#[cfg_attr(feature = "ast_serde", derive(serde::Serialize, serde::Deserialize))]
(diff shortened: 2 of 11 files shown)


schungx:
If I remember there are problems with it.  First of all, some AST structures that holds a `Dynamic` cannot be serialized.

Then I remember the `switch` statement block doesn't work.

Function hashes are volatile from run to run so they are also volatile unless you specifically compile for fixed hash keys.

All in all, your deserialized scripts may not run correctly. Not to mention you'll have huge issues with versioning.

schungx:
If I remember there are problems with it.  First of all, some AST structures that holds a `Dynamic` cannot be serialized.

Then I remember the `switch` statement block doesn't work.

Function hashes are volatile from run to run so they are also volatile unless you specifically compile for fixed hash keys.

All in all, your deserialized scripts may not run correctly. Not to mention you'll have huge issues with versioning.

IMHO there are ways to protect the script text from prying eyes just like any other secrets. Better use those tricks.

fluxlinkage:
> If I remember there are problems with it. First of all, some AST structures that holds a `Dynamic` cannot be serialized.

They can be serialized, as `Dynamic` can be serialized.

> Then I remember the `switch` statement block doesn't work.

The implement is different from https://github.com/schungx/rhai/tree/ast_serde. In this fork, all kinds of statements work, including `switch`.

> Function hashes are volatile from run to run so they are also volatile unless you specifically compile for fixed hash keys.
> 
> All in all, your deserialized scripts may not run correctly. It must be tested exhaustively.
> 
> Not to mention you'll have huge issues with versioning since a single change in AST structs will invalidate your entire scripts base.

Yes, you are right.

schungx:
> They can be serialized, as `Dynamic` can be serialized.

Not really...  if it is a variant, then it only serializes as the type name.

PR #944, opened 2024-12-31
fix extra newlines in raw strings

benatkin:
fixes #941 

I wonder how it decides to call `pos.advance()` and `pos.newline()`


benatkin:
It looks like it's fine calling `pos.advance()` and still calling `pos.new_line()`. It corrects from being one past the last character of the line to being the first character of the new line - it doesn't really advance twice, though it wouldn't make a difference if it didn't advance when the character is a newline. AFAIK this PR is ready.

Issue #946, opened 2025-01-04
Fails to build on some no_std targets with `sync`


stargazing-dino:
related: https://github.com/rhaiscript/rhai/pull/924

When building for no_std targets and using the `sync` feature, we get bizarre errors like this:

```sh
error[E0277]: the trait bound `RwLock<StringsInterner>: core::convert::From<StringsInterner>` is not satisfied
  --> /Users/katzen/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rhai-1.19.0/src/api/mod.rs:66:68
   |
66 |             self.interned_strings = Some(StringsInterner::new(max).into());
   |                                                                    ^^^^ the trait `core::convert::From<StringsInterner>` is not implemented for `RwLock<StringsInterner>`, which is required by `StringsInterner: core::convert::Into<_>`
   |
   = note: required for `StringsInterner` to implement `core::convert::Into<RwLock<StringsInterner>>`
```

Here's a [minimal reproducible example](https://github.com/stargazing-dino/test_rhai_embedded)

I tried pointing it to my local fork of Rhai to further debug but that's when I got what seems to be the more appropriate errors:

<details>
  <summary>Full Error</summary>

```
❯ cargo build
    Blocking waiting for file lock on package cache
    Updating git repository `https://gitlab.com/jD91mZM2/no-std-compat`
    Updating crates.io index
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on build directory
   Compiling spin v0.7.1
   Compiling rhai v1.20.1 (/Users/katzen/Documents/rhai)
   Compiling rhai_codegen v2.2.0 (/Users/katzen/Documents/rhai/codegen)
   Compiling no-std-compat v0.4.1 (https://gitlab.com/jD91mZM2/no-std-compat#47a5dfb6)
error[E0432]: unresolved import `std::sync::Arc`
  --> /Users/katzen/Documents/rhai/src/func/native.rs:36:9
   |
36 | pub use std::sync::Arc as Shared;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^ no `Arc` in `generated::sync`
   |
help: consider importing this struct instead
   |
36 | pub use alloc::sync::Arc as Shared;
   |         ~~~~~~~~~~~~~~~~~~~~~~~~~~

error[E0432]: unresolved import `crate::Shared`
[...]

PR #947, opened 2025-01-04
Use alloc for `Shared` type

stargazing-dino:
fixes https://github.com/rhaiscript/rhai/issues/946

This was smallest possible changeset that got rhai compiling and building for my test setup with the two features `sync` and `no_std`, which is odd because I see a lot more references to `std` in `native.rs` but who knows

I still got two compilation warnings but I couldn't get them to go with any combination of cfg's

<details>
  <summary>Warnings</summary>

```
➜ cargo build
   Compiling rhai v1.20.1 (/Users/katzen/Documents/rhai)
warning: unused import: `num_traits::Float`
  --> /Users/katzen/Documents/rhai/src/func/call.rs:31:5
   |
31 | use num_traits::Float;
   |     ^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `num_traits::Float`
   --> /Users/katzen/Documents/rhai/src/packages/logic.rs:231:9
    |
231 |     use num_traits::Float;
    |         ^^^^^^^^^^^^^^^^^

warning: `rhai` (lib) generated 2 warnings
   Compiling test_rhai_embedded v0.1.0 (/Users/katzen/Documents/test_rhai_embedded)
```
</details>

I wonder if this fixes things for you too @jeromegn

### DIFF
--- a/src/func/native.rs
+++ b/src/func/native.rs
@@ -30,10 +30,10 @@ impl<T> SendSync for T {}
 
 /// Immutable reference-counted container.
 #[cfg(not(feature = "sync"))]
-pub use std::rc::Rc as Shared;
+pub use alloc::rc::Rc as Shared;
 /// Immutable reference-counted container.
 #[cfg(feature = "sync")]
-pub use std::sync::Arc as Shared;
+pub use alloc::sync::Arc as Shared;
 
 /// Synchronized shared object.
 #[cfg(not(feature = "sync"))]
--- a/src/func/native.rs
+++ b/src/func/native.rs
@@ -30,9 +30,16 @@ impl<T> SendSync for T {}
 
 /// Immutable reference-counted container.
 #[cfg(not(feature = "sync"))]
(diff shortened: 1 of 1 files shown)


schungx:
Rhai uses a crate called `no_std_compat` or something similar which remaps most `std` namespaces into equivalent no-std ones.

I believe `std::sync::Arc` is already mapped to the equivalent `alloc` version...

However there may be naming conflicts that cause the mapping to go haywire. As I said I cannot reproduce it so it can easily be a weird combination of flags etc.

stargazing-dino:
> Rhai uses a crate called no_std_compat

Ah! at one point I knew this but forgot it existed.

Also, I added the comments and todos. Let me know if they're okay or need rephrasing

schungx:
Thanks!

schungx:
@stargazing-dino seems like it is failing CI on normal builds. Probably needs to `extern alloc` or something.

PR #948, opened 2025-01-05
Rm feature flag on `extern crate alloc`

stargazing-dino:
Try to fix the failing build here: https://github.com/rhaiscript/rhai/actions/runs/12616498242/job/35157801445

This is a very blunt solution but might be a good stop gap.


schungx:
I'm faster: https://github.com/rhaiscript/rhai/pull/949  LOL

PR #968, opened 2025-03-20
Make building with time support on wasm optional

TrueDoctor:


schungx:
I think there is a `no_time` feature...

TrueDoctor:
This still pulls in the crate dependency

schungx:
It pulls it in but does not use it. I am not sure there is a way in `Cargo.toml` to explicitly omit a dependency when a feature is set. 

Do you have reservations on pulling the crate in?

TrueDoctor:
Our CI fails with a deprecation warning, but I guess we could also add `instant` as an exception. That is just a slightly imperfect solution because we are then no longer warned about adding the dep elsewhere

schungx:
I just hesitate to do major surgery here as itay break some build combinations.

TrueDoctor:
This shouldn't really break anything because previously, compiling without either the `wasm-bindgen` or `std-web` feature always produced a build error. I've moved the `get_random` dependency to non-optional target dependency because from what I could tell, it is not actually `wasm-bindgen` or `std-web` specific. 
T = Target wasm{32/64}-unknown-unknown, W = `wasm-bindgen` | `std-web` feature used
Before:
| W | T |Result|
|---|------|----|
|  ❌ | ❌ | Compiles fine|
| ✅ | ❌ | Compile error invalid target |
| ❌ | ✅ | Compile error must choose W |
| ✅ | ✅ | Compiles fine |

After:
| W | T | `no_time` |Result|
|---|------|----|---|
|  ❌ | ❌ | `_` | Compiles fine|
| ✅ | ❌ | `_` | Compile error invalid target |
| ❌ | ✅ | ❌ | Compile error undeclared crate or module `instant` |
| ❌ | ✅ | ✅ | Compiles fine |
| ✅ | ✅ | `_`  | Compiles fine |

The Error message in case the `no_time` feature is not used should probably be improved

schungx:
Would you mind putting in the relevant combinations in CI? 

I'll merge it afterwards.

schungx:
OK, it seems that the following new checks is failing CI:
```toml
#[cfg(target_family = "wasm")]
#[cfg(not(any(feature = "no_time", feature = "wasm-bindgen", feature = "stdweb")))]
compile_error!("WASM targets have to use either the `wasm-bindgen`, `stdweb` or `no_time` feature");
```

I think it can be removed because it should be possible to build for WASI?

TrueDoctor:
Ah, I see that makes sense. Should I just revert my laestest commit then?

schungx:
@TrueDoctor   OK, it seems the following CI has failed:

```sh
cargo build --target wasm32-unknown-unknown --no-default-features
```

Looks like `Instant` is not included under this scenario, which is a _raw_ WASM build.

TrueDoctor:
Yeah, but shouldn't that have also failed before then?

schungx:
> Yeah, but shouldn't that have also failed before then?

I don't think so...

PR #979, opened 2025-04-18
Embed environment into FnPtr.

schungx:


thalesfragoso:
I'm getting "variable not found" errors when accessing constants in a rhai function called from rust through a `FnPtr`. The constants are in the same module, but it seems this PR (and the previous one) is still needed for my situation to work, is my assumption correct?

thalesfragoso:
Hmm, I tried my project with main but I still get the same error. I can try to prepare a minimal repro if this isn't a known issue.

schungx:
Yes, a min repo would help greatly in identifying the problem.

PR #982, opened 2025-04-27
added From<T> implementation for Option<T>

lukasx999:


schungx:
Wouldn't it be better to use `x.map_or(Dynamic::UNIT, From::from)`?

lukasx999:
Thank you, I made the according changes.
I tried to make the trait implementation generic over the value contained inside the Option, but it seems that's conflicting with the implementation of `From<ImmutableString>`.

schungx:
What I mean is...  why implement the trait when the user can simply use `map_or` to provide a default value that he/she wants instead of having it baked in?

lukasx999:
I suppose because its more convenient, and I think most users would expect `Option::None` to be converted to `()`.

schungx:
> I suppose because its more convenient, and I think most users would expect `Option::None` to be converted to `()`.

True, but then shouldn't we have `From<Option<...>>` for other built-in types as well?

lukasx999:
That's what I tried to do, but as mentioned it didn't quite work out.
> I tried to make the trait implementation generic over the value contained inside the Option, but it seems that's conflicting with the implementation of From<ImmutableString>.

schungx:
Yes you'd have to implement each type separately.

schungx:
I'm marking this as `wont_fix` as it'll probably need an exhaustive implementation for all the basic types, and I'm not very sure it would be significantly better than the alternative using `map_or(Dynamic::UNIT, Into::into)`.

Issue #984, opened 2025-04-29
Frame before call to rust a function missing


rhizoome:
**HEAVY EDIT** after I understood a crucial detail.

The frame before calling to a rust function is missing. I can inject the frame in the rust function called, but there are two problems:

- Source is not set in `NativeCallContext`
- When I have a method of a struct it is difficult to get `NativeCallContext`

(I was able to get a `NativeCallContext` for methods using a wrapper written in rhai.)

**My wish:** To have the call to `my_err` automatically in the source chain, so I could avoid both problems.  However, I suspect I am misunderstanding something.

```rust
use rhai::{Engine, EvalAltResult, FuncRegistration, NativeCallContext, Position};

fn my_err(ctx: NativeCallContext) -> Result<(), Box<EvalAltResult>> {
    // Name is set
    println!("fn_name: {:?}", ctx.fn_name());
    // Source is missing
    println!("source: {:?}", ctx.source());
    // Position set
    println!("postiion: {:?}", ctx.position());
    // I'd like to emulate the rhai error pattern for the rust function, so users can send me traces,
    // of course here we did not call rust_function
    let src = EvalAltResult::ErrorInFunctionCall(
        "rust_function".into(),
        file!().into(),
        "a rust error happened".into(),
        Position::new(line!() as u16, 0),
    );
    EvalAltResult::ErrorInFunctionCall(
        ctx.fn_name().into(),
        ctx.source().unwrap_or("").into(),
        src.into(),
        ctx.position(),
    )
    .into()
}

fn main() {
    let mut engine = Engine::new();
    engine.set_optimization_level(rhai::OptimizationLevel::None);
    FuncRegistration::new("my_err")
        .in_global_namespace()
        .register_into_engine(&mut engine, my_err);
    let mut ast = engine
        .compile(
            r#"
fn test_inner() {
    my_err();
}

fn test_other() {
    test_inner();
}

test_other();
    "#,
        )
        .unwrap();
    ast.set_source("inline");
    let err = engine.run_ast(&ast);
    println!("err: {}", err.err().unwrap());
}
```

Output

```
fn_name: "my_err"
source: None
postiion: 3:5
err: Runtime error: a rust error happened
in call to function 'rust_function' @ 'src/main.rs' (line 17, position 0)
in call to function 'my_err' (line 3, position 5)
in call to function 'test_inner' @ 'inline' (line 7, position 5)
in call to function 'test_other' @ 'inline' (line 10, position 1)
```

[...]

PR #986, opened 2025-05-02
Use caller source instead of callee source

### LINKED ISSUE #985: AST source is always `None` in `NativeCallContext<'_>`
Steps to reproduce:

```rs
let mut e = rhai::Engine::new();
e.register_fn("print_source", |ctx: rhai::NativeCallContext<'_>| {
    println!("{:?}", ctx.source());
    println!("{:?}", ctx.position());
});
let mut ast = e.compile_expression("print_source()").unwrap();
ast.set_source("my_source");
e.eval_ast::<()>(&ast).unwrap();
```

Expected output:
```
Some("my_source")
1:1
```

Actual output:
```
None
1:1
```

This happens in v1.21.0

HactarCE:
Currently calling a function uses the source of the callee but the position of the caller. This seems strange. I think that the source of the caller should be used instead, although I'm not exactly sure what the expected behavior is.

This fixes #985

### DIFF
--- a/src/func/call.rs
+++ b/src/func/call.rs
@@ -397,7 +397,7 @@ impl Engine {
             // Run external function
             let context = func
                 .has_context()
-                .then(|| (self, name, source.as_deref(), &*global, pos).into());
+                .then(|| (self, name, global.source(), &*global, pos).into());
 
             let mut _result = match func {
                 // If function is not pure, there must be at least one argument
--- a/src/func/call.rs
+++ b/src/func/call.rs
@@ -355,7 +355,7 @@ impl Engine {
         let a = Some(&mut *args);
         let func = self.resolve_fn(global, caches, local_entry, op_token, hash, a, true);
 
-        if let Some(FnResolutionCacheEntry { func, source }) = func {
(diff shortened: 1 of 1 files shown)


HactarCE:
There may be other places that should be updated similarly; I'm not sure.

rhizoome:
@HactarCE I do not understand these things completely, actually I am very confused: #984. Just before you spend too much time on this: NativeCallContext might need another field called `call_source`, because `source` is not meant to show you the call source, it is as a cache key.

See here: https://github.com/rhaiscript/rhai/issues/984#issuecomment-2846046114

> Well, just some background regarding how source came into being.
> 
> Firstly we made the ability to load external scripts as modules.
> 
> Then we need to figure out where each function is (i.e. which file) in order to debug with line numbers.
> 
> Therefore source was added to track where a function is defined in order to make sense of line numbers.
> 
> The source was never intended for which file the function was called from. Maybe as you say a new field call_source...
> 
> But then it'd be easy for you to get it from the error itself -- just look at the source of the caller function.

schungx:
Or rename `source` to `func_source`?

schungx:
I'll close this PR because I'll put in a new one that adds the `call_source` field.

PR #987, opened 2025-05-02
Don't pull in unneeded dependencies when enabling features

HactarCE:
e.g., enabling the `send` feature should not enable the dependency `no-std-compat` unless `no-std` is already enabled. The `?/` syntax in `Cargo.toml` handles this correctly.

I also replaced `https://gitlab.com/jD91mZM2/no-std-compat` with `https://gitlab.com/jD91mZM2/no-std-compat.git`, which works around rust-lang/cargo#15481


schungx:
The question mark syntax... We need to be sure it is supported in the MSRV compiler...

PR #999, opened 2025-06-22
Make CustomType macro inherit struct's generic parameters

ProphetOSpam:
Allows using the `CustomType` proc_macro with a struct with generic parameters

e.g.:
```rust
#[derive(Clone, CustomType)]
struct Foo<T> {
    x: T,
}
```


schungx:
Merged but a few tests failed CI under the `sync` feature. 

I suppose different struct definition needs to be used under `sync` with T: Send + Synf`.

schungx:
Turns out it is more involved to handle `sync`, which requires adding a new feature flag to `codegen`.  Therefore, I'll upload a PR for it.

schungx:
Yes it is currently UB if the `&str` is in the type itself.

PR #1006, opened 2025-07-07
Use `web-time` crate instead of unmaintained `instant` crate

### LINKED ISSUE #891: Rhai depends on unmaintained crate "instant"

therealprof:
The `instant` maintainer recommends switching to `web-time` instead. This also replaces #968 (from which I've shamelessly stolen the getrandom change) and fixes #891.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,6 @@ serde = { version = "1.0.136", default-features = false, features = ["derive", "
 serde_json = { version = "1.0.45", default-features = false, features = ["alloc"], optional = true }
 unicode-xid = { version = "0.2.0", default-features = false, optional = true }
 rust_decimal = { version = "1.24.0", default-features = false, features = ["maths"], optional = true }
-getrandom = { version = "0.2.7", optional = true }
 rustyline = { version = "15.0.0", optional = true }
 document-features = { version = "0.2.0", optional = true }
 arbitrary = { version = "1.3.2", optional = true, features = ["derive"] }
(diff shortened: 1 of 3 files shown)


schungx:
This breaks code because you removed the `wasm-bindgen` and `stdweb` features. 

Also, is it good to always include `js` for wasm? Some wasm builds may not have JavaScript available.

therealprof:
Would you prefer blind feature flags or removing them from CI?

> Also, is it good to always include js for wasm? Some wasm builds may not have JavaScript available.

Honestly, I have no idea what the use case for the `wasm32-unknown-unknown` target even is nowadays... Free standing executors will most certainly use WASI and as far as I'm aware there's no way of using WASM in the browser without JS. The `Cargo.toml` could probably even be tighter to only include `web-time` for `wasm32-unknown-unknown` rather than all `wasm32` targets...

schungx:
> Would you prefer blind feature flags or removing them from CI?

Well, removing them would be breaking, although we can keep an empty feature just for the sake of compatibility...  I'm not hung up either way, just not break code.

> Honestly, I have no idea what the use case for the `wasm32-unknown-unknown` target even is nowadays...

I remember some user before having his own execution environment that is not a browser and not wasi...  That is why `wasm32-unknown-unknown` is full of `unknown`'s...  Otherwise it would probably be specified as `browser` or `js`...

therealprof:
I already added the flags back. The current build failure is also on `main` and comes from unpinned `ahash` pulling in the incompatible `getrandom` `0.3.3`:
```
# cargo tree --invert getrandom@0.3.3
getrandom v0.3.3
└── ahash v0.8.12
    └── rhai v1.22.2 (/Users/egger/OSS/rhai)
```

schungx:
> With `ahash` pinned to `0.8.4`, this PR builds just fine:

Yes, I know that.  `getrandom` introduces a breaking change that is breaking a lot of crates.

My point is that `web-time` does not seem to support `stdweb` which is supported by Rhai.  That would be breaking... 

Maybe the solution will be to introduce _yet_ another feature flag to switch to `web-time`...

schungx:
Do we get a confirmation whether it supports `no-std`?

Had been a while.  Let me find some time to redo this. 

There are changes made to `getrandom` and so this needs to be reworked anyway.

schungx:
I have merged https://github.com/rhaiscript/rhai/pull/1063 which switches to `web-time` and removes `instant` dependency.

Closing this PR.

Thanks for everyone for your comments.

PR #1007, opened 2025-07-07
Drive-by changes nagged about by clippy

therealprof:


src/func/builtin.rs
schungx, line 1: We have to consider the fact that, in the future, we may add stuff after the `match` statement, and that would require adding `return`'s to all the short-circuit branches.

I believe those `return`'s are there because originally there were stuff after the `match`.

It might be better to silence clippy here instead.

therealprof: Wouldn't it make more to refactor the code if actually needed rather than cripple readability just in case it might be? I'm fine backing out the changes but I don't quite see the point why code needs to be terrible to read now.

schungx:
> Fixed. It doesn't really help that many unittests fail when compiled with `only_i32`. I did check that the same unittests fail before and after the change, though.

You'd also need to back out the tests with `MAX_USIZE_INT` etc. because they are required for 32-bit.

schungx:
```rust
// Original
#[allow(clippy::cast_sign_loss, clippy::cast_possible_truncation)]
let len = len.min(MAX_USIZE_INT) as usize;

// New
let len = len as usize;
```

The new version is problematic.  It currently does not cause an CI failure simply because our CI does not fully test edge cases.  It'd probably prop up during fuzzing.

Consider a 32-bit CPU where `usize` is 32-bit.  `len` is `i64` which is 64-bit.

Casting `len` as `usize` truncates.  However, the clippy lint is turned off here so you don't see it.

schungx:
> `try_from` is certainly a good idea, if error handling is done right (rather than e.g. simply unwrapping).

Yup, and this is what I'm doing.  I'm almost done, but still need to comb through carefully just in case I make a mistake.

Maybe it'd be better for you to wait for my commit and then rebase off that.  There will be quite a number of changes to change all implicit castings to `try_from`.

schungx:
@therealprof   Please try to rebase off the latest drop.  I have replaced most non-trivial castings with `try_from` which should make Rhai much safer.

PR #1018, opened 2025-08-18
Fix buildscript: Don't write to source tree

matthiasbeyer:
This patch rewrites the buildscript to use OUT_DIR instead of writing to the source tree, which is the way to go.

Besides that we also replace the File::open and File::create shenanigans with the std::fs shortcuts provided by the standard library, to shorten the whole thing.

---

This one is a blocker for my work with rhai, so a patch release including this patch would be awesome! :heart: :laughing:

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,20 +1,15 @@
-use std::{
-    env,
-    fs::File,
-    io::{Read, Write},
-};
+use std::env;
 
 fn main() {
+    let out_dir = env::var_os("OUT_DIR").unwrap();
+
     // Tell Cargo that if the given environment variable changes, to rerun this build script.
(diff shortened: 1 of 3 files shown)


schungx:
Well, turns out there is a problem with `OUT_DIR` afterall. It seems that, when loading a crate from `crates.io`, `OUT_DIR` is not defined when used outside of `build.rs`.

It caused loading Rhai from `crates.io` to fail,so I have reverted this change. 

The Rust docs also suggests writing directly to the source tree under this situation, so I am not sure if we can avoid it.

matthiasbeyer:
Writing to the source tree does not work in nix builds.

PR #1027, opened 2025-09-06
Optimize Script and Method Calls

djgaven588:
A bit ago I spent some time optimizing Rhai for my project.

These changes helped out a lot for my use case, and are mostly focused on calling lots of scripts, and those scripts calling lots of methods.

Most of these changes involve reducing function calls (map_or_else for instance), using hashbrown always, or preventing excessive iteration (drains).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ rhai_codegen = { version = "3.1.0", path = "codegen" }
 
 no-std-compat = { git = "https://gitlab.com/jD91mZM2/no-std-compat.git", version = "0.4.1", default-features = false, features = ["alloc"], optional = true }
 libm = { version = "0.2.0", default-features = false, optional = true }
-hashbrown = { version = "0.15.0", optional = true }
+hashbrown = { version = "0.15.0" }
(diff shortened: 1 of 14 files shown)


schungx:
Thanks for the contribution!  There seem to be conflicts.  Can you resolve them?

schungx:
Also, I see that you have replaced `map_or` etc. with `if let`, and replaced iterators with `for` loops.  Theoretically speaking, these two styles should compile down to very similar machine codes, under release builds with LTO.  For example, `map_or` etc. are usually inlined and then turned into essentially `if let`.  Most iterators are replaced with their equivalent implementation of `for` loops.  That's what I found most when doing Rust.

Do you have benchmarks to show clear improvements in runtime?

Also can you elaborate a bit on what you intend to do with `hashbrown`?

schungx:
Closing this for now.  Feel free to reopen later.

djgaven588:
> Closing this for now. Feel free to reopen later.

:+1: All good, I've been putting my time into a new programming language I built (Hail) which solves this by moving to a stack machine so I don't think I'll be spending any more time on this.

While I may have outgrown it over the course of a year and a half, I thank you greatly for your time spent on Rhai!

PR #1028, opened 2025-09-12
Expose type aliases that are used in pub interfaces

matthiasbeyer:
These type aliases were private before.
Using them is not possible.

This patch exposes them, so users do not have to work around that limitation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -131,13 +131,13 @@ mod types;
 /// Error encountered when parsing a script.
 type PERR = ParseErrorType;
 /// Evaluation result.
-type ERR = EvalAltResult;
+pub type ERR = EvalAltResult;
 /// General evaluation error for Rhai scripts.
-type RhaiError = Box<ERR>;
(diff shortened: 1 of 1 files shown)


schungx:
My thinking is... it would be easy for someone to define their own shorthand types.  I'm not sure the shorthands, as they are named right now, are the best names.  If we open them to public, we can no longer rename them easily in the future.  Therefore, we might as well spend some more time pondering alternative names before we settle on a set...

Do you feel `PERR`, `RhaiResultOf` etc. are good?

schungx:
> I _guess_ there might be some tooling that can be used in CI to ensure some stuff is not visible in public interfaces, but I forgot what the tool was.

`rustdoc` generated docs will expand the type aliases if they are not public. I never paid enough attention to rustanalyzer...

But you can press F12 to jump to the actual definition?

schungx:
I'm going to close this for the time being. Feel free to reopen.

Issue #1033, opened 2025-10-01
rkyv over serde / as a feature flag?


wheregmis:
RKYV's benches compared to serde is insane. Are there any plans to use rkyv over serde or a seperate feature with rikyv  in near future?

schungx:
Not very familiar with it but I don't see why not... 

Does it coexist with serde well? If so we can simply gate the feature under a feature flag.

schungx:
Can you make a PR? As I don't use it and not too familiar.

wheregmis:
> Can you make a PR? As I don't use it and not too familiar.

Sure, as i am not also fully aware of the rhai side of the things. I will try to create a draft PR as an experiement. Where we can maybe work together to see if its feasible or can gain performance increment and plan it together.

PR #1034, opened 2025-10-03
Experimental Draft: Rkyv

wheregmis:
#1033

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,23 +18,50 @@ keywords = ["scripting", "scripting-engine", "scripting-language", "embedded"]
 categories = ["no-std", "embedded", "wasm", "parser-implementations"]
 
 [dependencies]
-smallvec = { version = "1.7.0", default-features = false, features = ["union", "const_new", "const_generics"] }
+smallvec = { version = "1.7.0", default-features = false, features = [
+    "union",
+    "const_new",
+    "const_generics",
+] }
 thin-vec = { version = "0.2.13", default-features = false }
-ahash = { version = "0.8.4", default-features = false, features = ["compile-time-rng"] }
+ahash = { version = "0.8.4", default-features = false, features = [
+    "compile-time-rng",
+] }
 num-traits = { version = "0.2.14", default-features = false }
-once_cell = { version = "1.20.1", default-features = false, features = ["race", "portable-atomic", "alloc"] }
+once_cell = { version = "1.20.1", default-features = false, features = [
+    "race",
+    "portable-atomic",
+    "alloc",
+] }
 bitflags = { version = "2.3.3", default-features = false }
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -18,36 +18,68 @@ keywords = ["scripting", "scripting-engine", "scripting-language", "embedded"]
 categories = ["no-std", "embedded", "wasm", "parser-implementations"]
 
 [dependencies]
-smallvec = { version = "1.7.0", default-features = false, features = ["union", "const_new", "const_generics"] }
```
schungx, line 21: See if you can avoid generating many changes due to auto-reformatting. This distracts from the actual changes in the file.

benches/rkyv.rs
```diff
@@ -0,0 +1,143 @@
+#![feature(test)]
```
schungx, line 1: Benches are run upon commit in my dev repo, so I'm not sure if we'll blow up GitHub by adding so many new benches. We can try. 

The benchmarks are not run in the official repo which is named `main`. I named mine `master` which will trigger running benchmarks and collecting performance data upon push.

src/rkyv/mod.rs
```diff
@@ -0,0 +1,93 @@
+//! _(rkyv)_ Zero-copy serialization and deserialization support for [`rkyv`](https://crates.io/crates/rkyv).
+//! Exported under the `rkyv` feature only.
+//!
+//! # Overview
+//!
+//! `rkyv` provides high-performance binary serialization with zero-copy deserialization.
+//! This is ideal for performance-critical scenarios like:
+//!
+//! * Script caching - Load compiled scripts 50-100x faster
+//! * State snapshots - Save/restore engine state with minimal overhead
```
schungx, line 10: I don't think rkyv will help in serializing a compiled AST or Engine run state here...

At least not at this point as we're only serializing `Dynamic` data. 

Actually this is an interesting idea... Can we serialize an AST such that it is zerocopy and can be passed directly into the Engine without recreating a full AST structure again?  

I'm not convinced this'll work but again, who am I to say it definitely cannot.

There is only value is the serialized AST can be used directly from the serialized bytes. If recreating the AST structure is needed, then reparsing the script text may not be too slow compared to a deserislization step.

wheregmis: Technically i think yes we can. But i am not sure how much benefits we can get over the complexity as we might have to implement Archieve trait to all AST types and i am finding hard time to make rkyv work with complex recursive structures. This is just my over the top answer as i havent gone in depth in this one. But i definately want to. Will keep you posted regarding this.

ImTheSquid:
Stale

yonas:
@wheregmis Any update on the PR?

PR #1047, opened 2025-10-25
fix: detect reserved keyword in functions params

TheAwiteb:
Check if the token is a keyword or reserved in the function parameters, if it's returns a `Reserved` error.

Ref: https://github.com/rhaiscript/rhai/discussions/1045


schungx:
Only one thing...  may not be a good idea to introduce a new error variant just for this case...  as existing code do not know how to deal with this.

Can we shoehorn it into an existing error?  Do you think?

How about `VariableExpected`?

schungx:
A parameter is itself a variable...  So technically speaking it is still kosher

PR #1050, opened 2025-11-08
Enhance optimizer to support constant maps

phsym:
This pull request enhances the optimizer to support constant propagation for map access expressions, enabling compile-time evaluation of property and index lookups when the map is a constant known at compile/optimization time. It also adds comprehensive tests to ensure these optimizations work as intended.

**Optimizer enhancements for constant map access:**

* Added support for optimizing property access (`var.property`) when `var` is a constant map, allowing direct replacement with the corresponding value at compile time.
* Added support for optimizing index access (`var[property]`) when `var` is a constant map and `property` is a string or a constant string variable, enabling constant folding for these expressions.
* Property access on constant map literals `#{a: 42}[property]` when `property` is a constant string variable is also optimized at compile time, removing the need for runtime lookup.

**Testing improvements:**

* Added a new test, `test_optimizer_const_map`, which verifies that constant map optimizations work for both property and index access patterns, including dynamic key access.

### DIFF
--- a/src/optimizer.rs
+++ b/src/optimizer.rs
@@ -954,6 +954,12 @@ fn optimize_expr(expr: &mut Expr, state: &mut OptimizerState, _chaining: bool) {
                 *expr = mem::take(&mut m.0).into_iter().find(|(x, ..)| x.name == p.2)
                             .map_or_else(|| Expr::Unit(*pos), |(.., mut expr)| { expr.set_position(*pos); expr });
             }
+            // var.property where var is a constant map
+            (Expr::Variable(v, _, pos) , Expr::Property(p, ..))  if state.propagate_constants && state.find_literal_constant(&v.1).map_or(false,Dynamic::is_map) => {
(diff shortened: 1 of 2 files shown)


src/optimizer.rs
```diff
@@ -1039,6 +1057,13 @@ fn optimize_expr(expr: &mut Expr, state: &mut OptimizerState, _chaining: bool) {
                 state.set_dirty();
                 *expr = Expr::CharConstant(s.chars().rev().nth(usize::try_from(i.unsigned_abs()).unwrap() - 1).unwrap(), *pos);
             }
+            // var[property] where var is a constant map variable
+            #[cfg(not(feature = "no_object"))]
+            (Expr::Variable(v, _, pos) , Expr::StringConstant(s, ..))  if state.propagate_constants && state.find_literal_constant(&v.1).is_some_and(Dynamic::is_map) => {
```
schungx: `is_some_and` is not supported in MSRV.

phsym: fixed

PR #1068, opened 2026-03-08
Bump getrandom to 0.4.x

haxtibal:
getrandom 0.4.0 was released on 2026-02-02. It can be used without code change.


schungx:
I believe it bumps the MSRV to 1.85...  I believe we might want to wait it out a bit to see the uptake of version 0.4.

PR #1071, opened 2026-03-16
improvement: update macros to not require manual imports

timokoesters:
This should make it easier for users of the library, because they don't have to add random `use` statements for all the things the macros need.

### DIFF
--- a/codegen/src/function.rs
+++ b/codegen/src/function.rs
@@ -625,7 +625,7 @@ impl ExportedFn {
         let mut dynamic_signature = self.signature.clone();
         dynamic_signature.ident = syn::Ident::new("dynamic_result_fn", Span::call_site());
         dynamic_signature.output = syn::parse2::<syn::ReturnType>(quote! {
-            -> RhaiResult
+            -> ::rhai::plugin::RhaiResult
         })
         .unwrap();
         let arguments: Vec<_> = dynamic_signature
@@ -651,7 +651,7 @@ impl ExportedFn {
                 #[doc(hidden)]
                 #[inline(always)]
                 pub #dynamic_signature {
-                    #name(#(#arguments),*).map(Dynamic::from)
+                    #name(#(#arguments),*).map(::rhai::Dynamic::from)
                 }
             }
         } else {
@@ -660,7 +660,7 @@ impl ExportedFn {
                 #[doc(hidden)]
                 #[inline(always)]
                 pub #dynamic_signature {
-                    Ok(Dynamic::from(#name(#(#arguments),*)))
+                    Ok(::rhai::Dynamic::from(#name(#(#arguments),*)))
                 }
             }
         }
@@ -717,7 +717,7 @@ impl ExportedFn {
                     input_type_names.push(arg_name);
                     input_type_exprs.push(
                         syn::parse2::<syn::Expr>(quote_spanned!(arg_type.span() =>
-                            TypeId::of::<#arg_type>()
+                            ::std::any::TypeId::of::<#arg_type>()
                         ))
                         .unwrap(),
                     );
@@ -754,7 +754,7 @@ impl ExportedFn {
                                 is_string = true;
                                 is_ref = true;
                                 quote_spanned!(arg_type.span().resolved_at(Span::call_site()) =>
-                                    mem::take(args[#i]).into_immutable_string().unwrap()
+                                    ::std::mem::take(args[#i]).into_immutable_string().unwrap()
                                 )
                             }
                             _ => unreachable!("why wasn't this found earlier!?"),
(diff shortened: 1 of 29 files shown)


timokoesters:
It could be made even more robust by adding the absolute path for Vec and Ok, but I think that's too messy to read and has little benefit

schungx:
My question is: what if the user pulls in Rhai under a different name?

schungx:
There are some CI tests not passing.  Can you also take care of those?  Thanks.

schungx:
Can you put in an attribute (e.g. `root` or `crate_name`) which defaults to `rhai`?

The test output failure I can fix.

codegen/src/lib.rs
```diff
@@ -201,7 +201,10 @@ pub fn export_fn(args: TokenStream, input: TokenStream) -> TokenStream {
         return e.to_compile_error().into();
     }
 
-    output.extend(function_def.generate());
+    // This function is deprecated and does not have custom root support
+    let root: Path = syn::parse_quote!(::rhai);
```
timokoesters, line 205: ::rhai is hard coded for this and a few more deprecated function.

codegen/src/function.rs
```diff
@@ -299,10 +293,8 @@ impl Parse for ExportedFn {
         if let Some(syn::FnArg::Typed(syn::PatType { ref ty, .. })) = fn_all.sig.inputs.first() {
             match flatten_type_groups(ty.as_ref()) {
                 syn::Type::Path(p)
-                    if p.path == context_type_path1
-                        || p.path == context_type_path1x
-                        || p.path == context_type_path2
-                        || p.path == context_type_path2x =>
+                    if p.path.segments.last().map(|s| s.ident.to_string())
+                        == Some("NativeCallContext".to_owned()) =>
```
timokoesters, line 297: This function is an implementation of Parser and I could not pass the custom root here. For checking NativeCallContext I instead only compare the last segment of the path.

timokoesters (commented):
I've implemented the root argument and added a test for it called one_arg_fn_with_context_custom_root

schungx:
That test failure is strange.  I'm looking into it, because I cannot reproduce it on my desktop, nor in my dev repo.

schungx:
Should we add some tests with `#[derive(CustomType(root="rhai"))]` and `#[export_module(root="rhai")]` just to test out the attribute?

I'll also update the Book to mention this attribute and remove the need to import.

codegen/src/function.rs
schungx: This seems infinitely better...

timokoesters:
Small correction, the syntax is 
```rust
#[derive(CustomType)]
#[rhai_type(root=oldrhai)]
struct MyCustomType {
}
```
I think more test would be useful. But you can rename rhai to something else, like `use rhai as oldrhai`.

Do you want me to add one derive test and one export_module test?

schungx:
> Small correction, the syntax is

Great!  I'll fix the Book.  I already added it in.  ;-D

But that would be a bit strange.  Is it not possible to do `#[derive(CustomType(root = old_rhai))]`?
 
> Do you want me to add one derive test and one export_module test?

That would be wonderful!

schungx:
> But that would be a bit strange. Is it not possible to do `#[derive(CustomType(root = old_rhai))]`?

I just searched and it seems that this is not possible.  So the `rhai_type` route is the only way.  Thus, please ignore this.

schungx:
> Do you want me to add one derive test and one export_module test?

Would you add the tests and I'll run CI one last time (baring the strange one that failed) and merge this.

codegen/tests/custom_root/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    let mut engine = oldrhai::Engine::new();
+    let m = oldrhai::exported_module!(advanced_math);
+    engine.register_static_module("Math::Advanced", m.into());
+
+    assert_eq!(
+        engine.eval::<oldrhai::FLOAT>(r#"let m = Math::Advanced::get_mystic_number();m"#)?,
+        42.0
+    );
+    panic!();
```
timokoesters: I added a panic here just to check if the CI actually tests this code. The CI should fail now.
If we then remove the panic, the CI should work and we can merge it.

schungx: Ok, running CI.

schungx:
CI does not fail at the panic, meaning that the tests are probably not run... 

Also the MSRV test failed because you're using 2024 edition. MSRV is 1.66 and you'd need to remove the edition in `Cargo.toml`.

timokoesters:
I pushed a commit to update the CI, please check if this is okay.

schungx:
OK they now panic.  This is great.

Remove the panic and I'll rerun CI once again.

schungx:
Missing feature `metadata` in `custom_root` test...

PR #1077, opened 2026-04-07
Bump codegen and remove unnecessary module imports

schungx:


src/func/plugin.rs
```diff
@@ -1,19 +1,13 @@
...
 use super::FnCallArgs;
-pub use super::RhaiFunc;
-pub use crate::{
-    Dynamic, Engine, EvalAltResult, FnAccess, FnNamespace, FuncRegistration, ImmutableString,
-    Module, NativeCallContext, Position,
-};
+use crate::NativeCallContext;
 #[cfg(feature = "no_std")]
 use std::prelude::v1::*;
-pub use std::{any::TypeId, mem};
```
thalesfragoso, line 11: Unfortunately, this was a breaking change for people that avoided the glob import (`rhai::plugin::*`) and in general, if we follow semver. This caused some breakage in one of my dependencies and I think it makes sense to revert it.

schungx: Thanks for the info. Can you make a PR? 

I suppose just keep those exports there until the next major version?

PR #1079, opened 2026-04-10
Fix compact_script dropping $raw$ custom-syntax bodies

yuvalrakavy:
## Summary

`Engine::compact_script` silently drops every character captured by custom syntax that uses the `$raw$` marker (registered via `register_custom_syntax_with_state_raw` or `register_custom_syntax_without_look_ahead_raw`). The compacted output contains the custom-syntax keyword and surrounding non-raw tokens, but the raw body is gone.

For example, a `grab { let x = 1; let y = 2; print(x + y); }` script compacts to `grab{` — body and closing brace both stripped. The compacted script then either fails to parse back or, worse, parses as a truncated no-op and only surfaces the corruption at a later recompile.

`$block$`, `$expr$`, `$ident$`, `$symbol$`, and `$token$` are all unaffected — only `$raw$` is broken. The plain `register_custom_syntax` API explicitly rejects `$raw$` (`src/api/custom_syntax.rs:254`), so this only affects the raw-variant registration APIs.

## Root cause

The `_char_mode` branch in `TokenIterator::next` (`src/tokenizer.rs`) pulls one character at a time via `self.stream.get_next()` and returns `Token::UnprocessedRawChar(ch)` early, **before** the normal token-compression block at the tail of `next()` runs. Raw characters therefore never reach the `compressed` buffer.

## Fix

Inside the `_char_mode` branch, before returning the raw char, append it verbatim to `control.compressed` when compression is active. Appending verbatim (rather than going through the normal identifier-boundary spacing path) is correct because raw content is opaque to Rhai — there is no grammar to strip whitespace from, and the plugin's parse function may be counting characters in its state machine. The surrounding non-raw tokens are still compacted normally, so the output remains smaller than the input.

Fix is ~10 lines in `src/tokenizer.rs`. No change to public API.

## Test plan

- [x] Added `test_compact_script_preserves_raw_custom_syntax_body` to `tests/custom_syntax.rs` — registers a minimal `grab { BODY }` raw syntax, compacts a script, and asserts that the body tokens still appear in the compacted form and that the compacted script round-trips through `compile`.
- [x] Full `cargo test --features internals` passes — 0 failures across all test binaries, including the existing `$raw$` tests (`test_custom_syntax_raw`, `test_custom_syntax_raw2`, `test_custom_syntax_raw_interpolation`).
- [x] Standalone minimal reproducer flips from exit 2 (bug) to exit 0 (fixed) with this patch applied.

## Background

Discovered while using a `timer NAME { BODY }` custom syntax in a downstream project (home-automation store server). The `timer` keyword uses `$raw$` to capture the body, and every stored script had its timer body dropped at validation time. The corruption only surfaced at cold load when the broken compact form failed to recompile, producing confusing "unexpected token" errors on scripts that had parsed cleanly moments earlier.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -2701,6 +2701,21 @@ impl Iterator for TokenIterator<'_> {
                     '\n' => self.pos.new_line(),
                     _ => self.pos.advance(),
                 }
+                // Compacting support for `$raw$` custom syntax: append the
+                // consumed raw character verbatim to the compressed buffer.
+                // Raw characters are not produced by `get_next_token` and
+                // carry no `last_token`, so the normal token-compression
+                // path at the tail of this function never sees them.
+                // Appending verbatim preserves the raw body exactly as the
+                // plugin authored it — which is correct, because raw
+                // content is opaque to Rhai (that is the whole point of
+                // `$raw$`).
+                if compress_script {
+                    let control = &mut *self.state.tokenizer_control.borrow_mut();
+                    if let Some(ref mut compressed) = control.compressed {
+                        compressed.push(ch);
+                    }
+                }
                 return Some((Token::UnprocessedRawChar(ch), pos));
             }
         }
--- a/tests/custom_syntax.rs
+++ b/tests/custom_syntax.rs
@@ -508,3 +508,77 @@ fn test_custom_syntax_raw_interpolation() {
         "SELECT  *   FROM   //table//  WHERE  id=? AND   value <= ?\n123\n42"
     );
 }
+
+/// Regression test: `Engine::compact_script` must preserve the body of custom
+/// syntax that uses `$raw$` character capture. Previously the raw-char path in
+/// `TokenIterator::next` returned early before the compression buffer was
+/// updated, silently dropping every character inside the raw capture.
+#[test]
+fn test_compact_script_preserves_raw_custom_syntax_body() {
+    use rhai::{Map, ParseError};
+
+    let mut engine = Engine::new();
+
+    // Register a trivial `$raw$` syntax: `grab { BODY }`. Parses by tracking
+    // brace depth over raw characters, stops at the matching `}`. Execution
+    // is a no-op — we only care about parsing and compaction.
+    engine.register_custom_syntax_without_look_ahead_raw(
+        "grab",
+        |symbols, state| {
+            if state.is_unit() {
+                *state = Dynamic::from(Map::new());
+            }
+            let mut map = state.write_lock::<Map>().unwrap();
+
+            let in_raw = map.get("in_raw").and_then(|v| v.as_bool().ok()).unwrap_or(false);
+
+            if in_raw {
+                let ch = symbols.last().and_then(|s| s.chars().next()).unwrap_or('\0');
+                let mut depth = map.get("depth").and_then(|v| v.as_int().ok()).unwrap_or(0);
+                match ch {
+                    '{' => {
+                        depth += 1;
+                        map.insert("depth".into(), Dynamic::from_int(depth));
+                    }
+                    '}' => {
+                        if depth == 0 {
+                            map.insert("in_raw".into(), Dynamic::from_bool(false));
+                            return Ok(None);
+                        }
+                        depth -= 1;
+                        map.insert("depth".into(), Dynamic::from_int(depth));
+                    }
+                    _ => {}
+                }
+                return Ok(Some("$raw$".into()));
+            }
+
+            match symbols.len() {
+                1 => Ok(Some("{".into())),
+                2 => {
+                    map.insert("in_raw".into(), Dynamic::from_bool(true));
+                    map.insert("depth".into(), Dynamic::from_int(0));
+                    Ok(Some("$raw$".into()))
+                }
+                n => Err(ParseError(Box::new(ParseErrorType::BadInput(LexError::UnexpectedInput(format!("unexpected len={}", n)))), Position::NONE)),
+            }
+        },
+        false,
+        |_ctx, _inputs, _state| Ok(Dynamic::UNIT),
+    );
+
+    let source = "grab { let x = 1; let y = 2; print(x + y); }";
+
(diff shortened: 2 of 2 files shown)


yuvalrakavy:
I have fixed the camera compilation error (with no_object). It should pass tests now

schungx:
Great catch! I already completely forgot `compact_script`.

There is still one test failing under `no_object`...

schungx:
Thanks, @yuvalrakavy for this PR.

Just purely out of curiosity, what are you doing with custom syntax?

yuvalrakavy:
Hi

Two things: I implemented timer functionality:

let name = “Hercules”;

timer timer_name after 5000 {
  print(`It took 5 seconds to ${name} to get here`);
}

Or

timer “visit” every 5000 {
  print(`Every 5 seconds ${name} is visiting here`);
}

And span for open telemetry:

span “sever-junk-request” {
  …
}

The application itself is object oriented state store. Class methods are
rhai scripts. This will be used as infrastructure for my house automation
system and will replace a a system that inspired the architecture that was
written 20 years ago and is running on Windows. The goal is to eventually
make it an open source software.


Yuval

schungx:
What a coincidence. My introduction to Rust started with me writing the next version of my home automation system which was developed 20 years ago on Windows with C# (I started with .NET since alpha version). I needed scripting to script events, and so I picked up Rhai, eventually became its maintainer. 

Would love to compare notes with you. 

I didn't go all the way to write my own object store. I used MSSQL and later moved to also support MySQL. What do you see as need to write your own object store inside of using an existing one?

yuvalrakavy:
Stephen — apologies for the slow reply. I wanted to give you a proper, informative answer rather than a hand-wave, and putting the write-up together took longer than I expected. What a lovely coincidence, by the way — I'd genuinely love to compare notes.
Funny parallel: around the same era you started writing your home automation system, I was picking up C# and writing — and still maintaining — control software for model trains. Different domain, same flavor of problem (typed object model, real-time events, hardware bridges, scripting for behavior), and a similarly long arc. The Store is in many ways the substrate I wish I'd had back then.
I actually wrote up the overview specifically with you in mind, as a first pass at the article we'll publish alongside the eventual open-source release. It's here:

https://gist.github.com/yuvalrakavy/cfb8a7169e029509e63da20663f26afb

The short version: the system is called the Store — a hierarchical, transactional object store in Rust, where schema, runtime state, automations, and UI logic all live in the same typed tree. Rhai is the scripting substrate everywhere: event handlers, typed store methods, instance/static methods, display formatters, timer bodies, MQTT topic handlers, scheduler conditions. Scripts appear inside an SDL (Schema Definition Language) file that also declares the class itself, so one file declares schema + behavior + reactivity.
A few Rhai-specific things you might find interesting:

on_missing_function is load-bearing for us. We don't pre-register class methods on the engine; instead, on every unresolved call we walk the target object's class chain at call time and dispatch to the appropriate fn / store fn. This is the hook I contributed upstream — it's what makes polymorphic, class-defined Rhai methods possible without forking the language. Thank you for merging it.
Custom timer syntax (the PR you just merged). It reads timer "name" every 1000 { ... }, plus after, after ... every, update, and kill. Implemented with $raw$ + a nested-brace state machine in the parser so the body is a real nested block. The @ in a name expands to self.id for per-instance identity. Scope variables persist across firings. I'd love your eyes on it when you have a moment.
Compile-once-cache-forever for every script; per-class modules merged at load; RhaiObject bridge so obj["Prop"], obj.children, obj.Method(...) all feel native; blocking-task adapters to bridge to Tokio without stalling.

The first application on top is HomeTouch — Lutron / DALI / DMX / HDL / CoolMaster drivers (each a separate MQTT bridge, never in-process), a sunset-aware scheduler, and touchscreens that are true thin clients streaming RFB from a server-side renderer fed by a live gRPC mirror of the object tree. The ancestor is Premise SYS from the early 2000s — which, from your description, sounds like it might be the same era of idea you're revisiting.
Would love to hear about your system. Happy to jump on a call whenever suits you.
Yuval

schungx:
Wow, that was intense.  Mine wasn't in the same League, although the devices we tackled were similar: Lutron, C-Bus, EIB, Midea (their earlier attempt to connect, I worked with the engineer himself), HDL, CoolMaster, and for me also the ELK security panel (designed an abstraction of a security system, but eventually only wrote for the ELK).  Interesting as HDL wasn't a very big name in the space at that time, so you started early with them (as did we).  Also streaming video (Hulu/Netflix didn't exist at that time) and infrared redirection (people still switch channels, Smart TV didn't exist).

I originally use C# and VB.NET as the scripting language, which was quite easy to do in .NET.  Moving to Rust, I had to find a new scripting language.  Lua was too hard to integrate at that point (who wants to spend days figuring out how to connect to an essentially C-based library instead of writing interesting code?), JavaScript I hate (had too much of it already, I was one of the very earliest adapters of TypeScript), and then I discovered Rhai.  It was still in a relatively early phase of development, but it was simple enough and easy to use enough.

My system was relatively basic, but it was already enough for me to start a business in the 2000's on it.  I still have clients paying for regular maintenance, although these days we don't do such systems anymore because cell phones are everywhere, and every electrical and electronic gadget has an App...  We are still developing v2 (actually version 3, there are two versions in C#) of the system in Rust, and incidentally choosing to use MQTT for the unified message bus.  Incidentally, my system worked exactly like you said: all devices and schema definitions (I called them "Zone" definitions, each room in a premise a zone) reside on the same type tree and can be inspected or written to (which triggers actions translated via drivers to the actual devices).  Scripts are used as triggers to tie the whole network together into a unified platform (although no atomic updates...  I deliberately used an API model which is fire-and-forget...  you poll for the result and continue when it happens or handle time-out).  So it seems we came around to the same solutions at around the same time.

Just out of curiosity...  what is the need for `$raw$` for the block?  Why not just use a `$block$` and be done with it?

Also, are you using `$raw$` for definition blocks (in another syntax)?

PR #1083, opened 2026-04-21
feat(internals): expose fn-resolution cache isolation on EvalContext

yuvalrakavy:
# Upstream PR: Expose fn-resolution cache isolation on EvalContext

> Ready to submit to `rhaiscript/rhai`. Source branch
> `yuvalrakavy/rhai:expose-fn-cache-isolation` (commit `0b0f4f19` on top
> of current `upstream/main` at `c3202b62`). Diff: 1 file, 34 insertions.

## Suggested PR title

```
feat(internals): expose fn-resolution cache isolation on EvalContext
```

## Suggested PR body

```markdown
## Summary

Forwards three existing internal `Caches` methods — `push_fn_resolution_cache`, `fn_resolution_caches_len`, `rewind_fn_resolution_caches` — through the public `EvalContext` API so external `on_missing_function` callbacks that manipulate `global.lib` can isolate their dispatch cache frame the same way the core evaluator already does internally.

Purely additive. 34 insertions, 0 deletions, 0 behavior change for existing callers. No feature gates required (can be gated behind `internals` if preferred — happy to adjust).

## Motivation

External callers using `on_missing_function` for custom method resolution commonly push a module onto `global.lib` and then call `ctx.call_fn_raw`, like this:

```rust
engine.on_missing_function(move |name, args, is_method_call, mut ctx| {
    let module = /* resolve which module defines `name` for this call */;
    let global = ctx.global_runtime_state_mut();
    global.lib.push(module);

    let result = ctx.call_fn_raw(name, true, false, args);

    ctx.global_runtime_state_mut().lib.pop();
    result.map(Some)
});
```

The fn-resolution cache is keyed by `(name, arg_types)` and does **not** invalidate when `global.lib` changes. So when the callback is re-entered from inside a called function (e.g. via a `super()`-style proxy that re-triggers dispatch with a different module), the second invocation hits the cached entry from the first call — routing nested dispatches to the wrong module's implementation.

The core evaluator already handles this at `src/eval/stmt.rs:64-95` and `src/func/script.rs:94-206`:

```rust
let orig_fn_resolution_caches_len = caches.fn_resolution_caches_len();
// ... manipulate global.lib + evaluate ...
caches.rewind_fn_resolution_caches(orig_fn_resolution_caches_len);
```

External hook callers cannot access the same pattern because the relevant methods sit on the crate-private `Caches` type, and `ctx.caches` is `pub(crate)`.

## Change

Three forwarding methods on `EvalContext`:

```rust
impl<'a, 's, 'ps, 'g, 'c, 't> EvalContext<'a, 's, 'ps, 'g, 'c, 't> {
    #[inline(always)]
    pub fn push_fn_resolution_cache(&mut self) {
        self.caches.push_fn_resolution_cache();
    }

    #[inline(always)]
    #[must_use]
    pub fn fn_resolution_caches_len(&self) -> usize {
        self.caches.fn_resolution_caches_len()
    }

    #[inline(always)]
    pub fn rewind_fn_resolution_caches(&mut self, len: usize) {
        self.caches.rewind_fn_resolution_caches(len);
    }
}
```

Usage from downstream callers:

```rust
engine.on_missing_function(move |name, args, is_method_call, mut ctx| {
    // ... resolve module ...
    let global = ctx.global_runtime_state_mut();
    global.lib.push(module);

    // Isolate this nested dispatch from any outer cache frame.
    let cache_len = ctx.fn_resolution_caches_len();
    ctx.push_fn_resolution_cache();

    let result = ctx.call_fn_raw(name, true, false, args);

    ctx.rewind_fn_resolution_caches(cache_len);
    ctx.global_runtime_state_mut().lib.pop();
    result.map(Some)
});
```

## Alternatives considered

1. **Clear the entire cache after every `call_fn_raw`** — penalizes every caller; the push/rewind pattern is scoped to the boundary that actually needs isolation.
2. **Use `call_fn_with_options` instead of `call_fn_raw`** — doesn't help; the cache issue is orthogonal to which call surface is used.
3. **Expose `ctx.caches` directly** — leaks the full `Caches` type and all its cache subsystems into the public API. Exposing just the three needed methods keeps the surface minimal and intent-revealing.

## Testing

Verified locally against `upstream/main + this commit` across the full upstream CI matrix from `.github/workflows/build.yml`:

- **All 22 `build` feature combos**: passed (counts range from 98 to 145 tests per combo depending on features):
  - default, `internals`, `debugging`, `metadata`, `serde`, `decimal`, `no_float+decimal`, `unicode-xid-ident`
  - kitchen-sinks: `sync+internals+serde+metadata+debugging`, `decimal+kitchen`, `unchecked+kitchen`, `no_position+kitchen`, `no_optimize+kitchen`, `no_float+kitchen`, `f32_float+kitchen`, `no_custom_syntax+kitchen`, `only_i64+kitchen`, `no_index+kitchen`, `no_object+kitchen`, `no_function+kitchen`, `no_module+kitchen`, `no_time+kitchen`, `no_closure+kitchen`
  - `only_i32+kitchen` (`--tests` only)
  - "all-negatives" combos (sync + 10× `no_*` features; and unchecked + 11× `no_*` features)
- **MSRV** (Rust 1.66.0, `--features decimal,metadata,serde,debugging`, with `Cargo.msrv.lock`): clean.
- **`no_std` build** (`--manifest-path=no_std/no_std_test/Cargo.toml --profile unix`, nightly): clean.
- **`cargo fmt --check`**: clean.

- Regression behavior verified in a downstream project that uses `on_missing_function` for custom method resolution with multi-level inheritance (three-level `super()`-style chains). Before this change, three-level chains return the most-derived class's implementation at every step; after, each level resolves correctly.

- **Dedicated regression test in `tests/`:** not included in this PR. An attempted minimal reproduction (via `set_native_fn` on two `Module::new` fixtures + two `call_fn_raw` invocations within a single `eval`) ran into a separate question about how `call_fn_raw` searches `global.lib` for native functions versus script functions — the downstream use case works because it holds compiled Rhai modules (via `get_script_fn`), not native-fn modules. Happy to draft a test using compiled script modules if you'd like to see a regression in-tree — it needs a small parse + `Module::eval_ast_as_new` fixture rather than `set_native_fn`, but is otherwise straightforward.

## Scope

This is a minimal public-API addition. No type changes, no signature changes, no behavior changes for existing callers. Just closing a gap between what the internal evaluator does and what external `on_missing_function` consumers can do.

## Backwards compatibility

- No existing methods changed.
- No feature gates added or modified.
- Works with / without `sync`, `internals`, `no_std`, etc. (see Testing above).
```

## Submission steps (for when CI verification completes)

1. Confirm full-matrix CI passed locally on `expose-fn-cache-isolation` (background job `b2vgzohgj`).
2. Open the GitHub PR against `rhaiscript/rhai:main` from `yuvalrakavy/rhai:expose-fn-cache-isolation`.
3. Use the title and body above (adjust wording to personal voice if preferred).
4. Monitor CI; be ready to respond to maintainer feedback (e.g. "please add a regression test" — pre-drafted below).

## Suggested regression test (stretch / optional — needs iteration)

A naive version using `set_native_fn` on two modules hit a stack
overflow because native-fn modules pushed onto `global.lib` don't
participate in `call_fn_raw`'s lookup the way script-fn modules do
(the downstream use case holds compiled Rhai modules via
`get_script_fn`, not native). The working form needs a small
compiled-script fixture rather than native fns. Shape:

```rust
// PSEUDO-OUTLINE — needs fixture compilation
let ast_a = engine.compile(r#"fn greet() { "A" }"#).unwrap();
let arc_a: Shared<Module> = Module::eval_ast_as_new(Scope::new(), &ast_a, &engine).unwrap().into();
let ast_b = engine.compile(r#"fn greet() { "B" }"#).unwrap();
let arc_b: Shared<Module> = Module::eval_ast_as_new(Scope::new(), &ast_b, &engine).unwrap().into();

// ... on_missing_function alternates pushing mod_a vs mod_b per call,
//     using push_fn_resolution_cache / rewind_fn_resolution_caches ...

// assert!(out == "A,B", "with cache isolation, second call sees mod_b");
```

If submitted for review the maintainer is likely to want this iterated to green before merge. Leaving it as a follow-up to avoid blocking the PR prep.

Prior attempt for reference (DOES NOT compile-or-run correctly, shown only to document the shape):

```rust
#[test]
#[allow(deprecated)]
fn test_on_missing_function_fn_cache_isolation() {
    use std::sync::atomic::{AtomicI64, Ordering};
    use std::sync::Arc;

    // Two modules both defining `greet()` — different return values.
    let mut mod_a = Module::new();
    mod_a.set_native_fn("greet", || Ok::<_, Box<EvalAltResult>>(Dynamic::from("A".to_string())));
    let arc_a: rhai::Shared<Module> = mod_a.into();

    let mut mod_b = Module::new();
    mod_b.set_native_fn("greet", || Ok::<_, Box<EvalAltResult>>(Dynamic::from("B".to_string())));
    let arc_b: rhai::Shared<Module> = mod_b.into();

    let counter = Arc::new(AtomicI64::new(0));
    let counter_clone = counter.clone();
    let arc_a_clone = arc_a.clone();
    let arc_b_clone = arc_b.clone();

    engine.on_missing_function(move |name, args, _is_method_call, mut ctx| {
        if name != "greet" { return Ok(None); }

        // First invocation pushes mod_a; second pushes mod_b.
        let which = counter_clone.fetch_add(1, Ordering::SeqCst);
        let module = if which == 0 { arc_a_clone.clone() } else { arc_b_clone.clone() };

        ctx.global_runtime_state_mut().lib.push(module);

        // Isolate cache frame so the second dispatch doesn't hit the first's cached entry.
        let cache_len = ctx.fn_resolution_caches_len();
        ctx.push_fn_resolution_cache();

        let result = ctx.call_fn_raw(name, true, false, args);

        ctx.rewind_fn_resolution_caches(cache_len);
        ctx.global_runtime_state_mut().lib.pop();

        result.map(Some)
    });

    // Two calls within one eval so they share GlobalRuntimeState.
    let out: String = engine.eval(r#"
        let x = 42;
        x.greet() + "," + x.greet()
    "#).unwrap();

    // Without cache isolation, both resolve to mod_a and output is "A,A".
    // With isolation, second call sees the freshly pushed mod_b.
    assert_eq!(out, "A,B",
        "cache isolation lets each on_missing_function invocation resolve against its current global.lib");
}
```

This test may need signature tweaks depending on Rhai's exact `set_native_fn` and `Shared<Module>` APIs on upstream; trying locally against the fork branch is a 5-minute check.

### DIFF
--- a/src/eval/eval_context.rs
+++ b/src/eval/eval_context.rs
@@ -364,6 +364,40 @@ impl<'a, 's, 'ps, 'g, 'c, 't> EvalContext<'a, 's, 'ps, 'g, 'c, 't> {
             false,
         )
     }
+
+    /// Push a fresh, empty function resolution cache onto the stack.
+    ///
+    /// This isolates subsequent function lookups from cached results obtained with a
+    /// different `global.lib` configuration (e.g. before an additional module was pushed).
+    /// Call [`rewind_fn_resolution_caches`][EvalContext::rewind_fn_resolution_caches] with the
+    /// value returned by [`fn_resolution_caches_len`][EvalContext::fn_resolution_caches_len]
+    /// to restore the previous cache state.
+    ///
+    /// Primarily useful inside `on_missing_function` callbacks that dynamically push modules
+    /// to `global.lib` and then call `call_fn_raw` — without this guard, the resolution cache
+    /// can return stale entries from before the lib change.
+    #[inline(always)]
+    pub fn push_fn_resolution_cache(&mut self) {
+        self.caches.push_fn_resolution_cache();
+    }
+
+    /// Return the current depth of the function resolution cache stack.
+    ///
+    /// Used together with [`rewind_fn_resolution_caches`][EvalContext::rewind_fn_resolution_caches]
+    /// to restore the cache stack after pushing a temporary frame.
+    #[inline(always)]
+    #[must_use]
+    pub fn fn_resolution_caches_len(&self) -> usize {
+        self.caches.fn_resolution_caches_len()
+    }
+
+    /// Rewind the function resolution cache stack to a previously saved depth.
+    ///
+    /// This pops all cache frames pushed since `len` was recorded.
+    #[inline(always)]
+    pub fn rewind_fn_resolution_caches(&mut self, len: usize) {
+        self.caches.rewind_fn_resolution_caches(len);
+    }
 }
 
 /// Call a function (native Rust or scripted) inside the [evaluation context][`EvalContext`].
--- a/tests/fn_resolution_cache_isolation.rs
+++ b/tests/fn_resolution_cache_isolation.rs
@@ -1,0 +1,116 @@
+//! Regression test for the fn-resolution cache bug that motivates the
+//! `EvalContext::push_fn_resolution_cache` / `fn_resolution_caches_len` /
+//! `rewind_fn_resolution_caches` API exposed by this PR.
+//!
+//! The bug: when an `on_missing_function` callback pushes a module onto
+//! `global.lib` and then calls `ctx.call_fn_raw`, the nested resolution is
+//! cached against whatever `global.lib` configuration is in effect at the
+//! time. Subsequent callback invocations that push a *different* module
+//! under the same `(name, arg_types)` hash will hit the cached (stale)
+//! entry and dispatch to the wrong module's function. Without a way to
+//! isolate the nested cache frame, external callers of `on_missing_function`
+//! cannot implement correct class-hierarchy dispatch, `parent_view()`/super
+//! walks, or any pattern that relies on swapping `global.lib` contents
+//! between calls of the same name.
+//!
+//! The fix: bracket `call_fn_raw` with `push_fn_resolution_cache` /
+//! `rewind_fn_resolution_caches` so each nested dispatch resolves in a
+//! fresh cache frame and discards its entries on exit. The primitives
+//! already exist on `Caches` internally (see `src/eval/stmt.rs:63-95` and
+//! `src/func/script.rs:94-206`); this PR just forwards them through
+//! `EvalContext` so `on_missing_function` callbacks can use them too.
+
+#![cfg(feature = "internals")]
+#![cfg(not(feature = "no_object"))]
+#![cfg(not(feature = "no_function"))]
+
+use rhai::{Array, Engine, Module, INT};
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+
+#[test]
+fn on_missing_function_isolates_nested_cache_frame() {
+    // Two modules, each registering `step(target, n) -> INT` under the same
+    // name and argument types. Same `(name, arg_types)` produces the same
+    // fn-resolution cache hash.
+    let mut module_a = Module::new();
+    module_a.set_native_fn(
+        "step",
+        |_target: &mut INT, n: INT| -> Result<INT, Box<rhai::EvalAltResult>> { Ok(n * 2) },
+    );
+    let module_a: Arc<Module> = Arc::new(module_a);
+
+    let mut module_b = Module::new();
+    module_b.set_native_fn(
+        "step",
+        |_target: &mut INT, n: INT| -> Result<INT, Box<rhai::EvalAltResult>> { Ok(n * 10) },
+    );
+    let module_b: Arc<Module> = Arc::new(module_b);
+
+    // Counter picks module_a for invocations 0-1, module_b for invocation 2+.
+    // Three invocations are needed to exercise the cache lifecycle: call 1
+    // sets the bloom filter bit (no dict entry yet), call 2 populates the
+    // dict against module_a, call 3 (with module_b pushed) must hit a FRESH
+    // resolution in its isolated frame — if isolation is missing it would
+    // hit the stale dict entry from call 2 and return module_a's result.
+    let call_idx = Arc::new(AtomicUsize::new(0));
+
+    let a = module_a.clone();
+    let b = module_b.clone();
+    let call_idx_clone = call_idx.clone();
+
+    let mut engine = Engine::new();
+
+    #[allow(deprecated)]
+    engine.on_missing_function(move |name, args, _is_method_call, mut ctx| {
+        if name != "step" {
+            return Ok(None);
+        }
+
+        let idx = call_idx_clone.fetch_add(1, Ordering::SeqCst);
+        let module = if idx < 2 { a.clone() } else { b.clone() };
+
+        ctx.global_runtime_state_mut().lib.push(module);
+
+        // Isolate the nested dispatch in a fresh cache frame. Without this,
+        // invocation 2 would cache the resolution against module_a, and
+        // invocation 3 would hit that stale cache entry after module_b had
+        // been pushed.
+        let orig_cache_len = ctx.fn_resolution_caches_len();
+        ctx.push_fn_resolution_cache();
+
+        let result = ctx.call_fn_raw(name, true, false, args);
+
+        ctx.rewind_fn_resolution_caches(orig_cache_len);
+        ctx.global_runtime_state_mut().lib.pop();
+
+        result.map(Some)
+    });
+
+    // Three method-style calls with identical name and argument types so
+    // they all hash to the same fn-resolution cache key.
+    let script = r#"
+        let x = 0;
+        [x.step(3), x.step(3), x.step(3)]
+    "#;
+
+    let results: Array = engine.eval(script).expect("eval must succeed");
+    let results: Vec<INT> = results
+        .into_iter()
+        .map(|v| v.as_int().expect("array element must be INT"))
+        .collect();
+
+    assert_eq!(
+        results,
+        vec![6, 6, 30],
+        "third invocation must dispatch to module_b (3*10=30), not a stale \
+         cached entry from module_a (3*2=6); without push/rewind_fn_resolution_cache \
+         the third element would be 6."
+    );
+
+    assert_eq!(
+        call_idx.load(Ordering::SeqCst),
+        3,
+        "on_missing_function must fire exactly three times"
+    );
+}


schungx:
Would you separate the changes to the `stderr` file from this PR? 

Currently there are two files changed.

Since there seems to be updates needed to make CI run with the latest Rust version, it might be better to do a separate PR. 

Also, I'm pondering whether there is a way to expose only one method for managing an extra caching later instead of three... 

Also, you need to be concerned that pushing a new cache makes precious cached calls to operators etc unavailable. That means your scripts would always run like the first time, clean but slow because every single function and operator call is resolved every time. It would be OK if you're only calling Rust... But if you're calling scripts at this point then the speed difference would be noticable.

An alternative would be to remove that cache entry after the call? That would mean somehow returning the hash after the function is being resolved. 

Second alternative is to pass in a flag to disable caching temporarily while keeping the cache intact. This might be an attractive solution that may have other uses. When other users need to handle highly polymorphic functions then it would be very handy

In conclusion, I think adding a `global.enable_add_to_caches(false);` would be more elegant... Although you'd need `global.add_to_caches()` to make sure we propagate the correct flag value from upstream.

schungx:
@yuvalrakavy I wonder whether you have tried the alternative flag to enable/disable caching...

yuvalrakavy:
Good catch to raise it — I wanted to make sure I wasn't missing an existing knob before answering. I went through `LangOptions`, `CallFnOptions`, and the `Engine` surface; the only `enable_cache` I could find is on `FileModuleResolver` (the module-loader cache — a different system). So I read your comment as a design proposal rather than pointing at an existing flag. If I missed one, please point me at it.

On the design itself: a simple on/off flag doesn't quite fit the specific bug this PR targets. The failing case is `parent_view()` chains — nested `on_missing_function` callbacks at different levels of an inheritance chain, each pushing its own class module onto `global.lib`. Those callbacks can be arbitrarily nested, so the "disable caching" state needs to be reentrant (ref-counted or stacked). Once you make it reentrant you've essentially reinvented push/rewind under a different name.

There's also a subtle perf inversion worth noting: "disable caching for the duration of the inner `call_fn_raw`" means every operator and helper-function lookup inside the inner function body re-resolves on every hit, including within loops. `push_fn_resolution_cache` + rewind keeps cache hits within the inner call (the new frame accumulates entries and they're reused within that scope) and only discards them on exit. For any non-trivial function body, frame-push is strictly faster than caching-disabled.

What I think your "one method instead of three" comment from the earlier review was pointing at is an RAII guard — a single method that returns a `Drop` guard, pushes on construction, rewinds on drop:

```rust
let _g = ctx.isolate_fn_resolution_cache();
let result = ctx.call_fn_raw(name, true, false, args);
// _g drops here, frame rewound automatically
```

Under the hood that's the same push/rewind pair, but with one public entry point and panic-safe pairing. Happy to make that the primary public API and demote the three primitives to advanced use (or drop them from the public surface entirely if you prefer). Let me know which shape you want.

On the cached-operators cost you flagged — thanks, that's a fair concern for downstream code, and we're addressing it on our side rather than asking rhai to widen its API. The approach is a small side-table maintained alongside our class-function module map that counts, per (name, arity), how many distinct classes define that signature. Dispatch then consults it in O(1) and only brackets `call_fn_raw` with push/rewind when the signature is defined by two or more classes. Monomorphic dispatches skip the isolation entirely and reuse the parent frame's cached operator/helper resolutions. That work is local to our callback and doesn't affect this PR.

And the stderr snapshot split is coming in its own PR.

schungx:
Ah, I take it for you to mean that, once a function call is made, it may further call a whole bunch of other functions within the same frame repeatedly (otherwise no use of caching)?

That's the only time when you want those calls cached. If there are no loops and no repeated calls to the same function, then actually you don't need to cache at all. 

Rhai actually does something like this. It never caches the first call of a function using a bloom filter and only caches on the second call. This is to avoid caching one-shot-wonders.

yuvalrakavy:
Thanks — your last message is actually the best framing of the trade-off, and I think it lands on push-frame for our case. The write-disabled flag would be cleaner as an API (and addresses the "one method" point), but your own observation nails why it's not enough: our class-fn bodies routinely call many other functions within the same frame, and some of them are hot loops. Write-disabled means those inner-body operator and helper lookups can't cache within the inner scope, so any call the body makes repeatedly re-resolves every iteration. Push-frame caches them in the isolated frame for reuse during that dispatch, then discards the frame on exit. For workloads like ours (schedulers, MQTT handlers, event cascades that iterate and do arithmetic inside class-fn bodies), that's a real win.

Meanwhile, on the cached-operator cost you flagged earlier: that's now addressed on our side. I shipped a side-table on our class-function module map that counts how many classes define each `(name, arity)` globally, and the `on_missing_function` callback only brackets `call_fn_raw` with push/rewind when the signature is polymorphic (two or more definers). Monomorphic class-fn dispatches — which is the overwhelming majority in practice — skip isolation entirely and reuse the parent frame's cached operator and helper resolutions. So the "run like the first time, clean but slow" concern is gone for the common case; only genuinely polymorphic dispatches (multi-level `parent_view()`, name-sharing across classes) pay the push/rewind cost, and those benefit from intra-body caching anyway.

An earlier chain-only variant of that gate was tried and reverted — rhai's resolution cache is keyed only by `(name, arg_types)`, so two unrelated classes with the same method name can contaminate each other within a single `eval()` if the gate only considers the current object's class chain. The global side-table avoids that by tracking definers across all classes, not just the chain. Happy to share the commit if useful context.

On the "one method instead of three" point from your earlier review — I'd like to address that by adding an RAII helper alongside the three primitives, something like:

```rust
let _g = ctx.isolate_fn_resolution_cache();
let result = ctx.call_fn_raw(name, true, false, args);
// _g drops here, frame rewound automatically
```

Under the hood it's still push + rewind. One public entry point for the common case, panic-safe pairing, and the three primitives remain available (or I can drop them from the public surface if you prefer them as internal-only implementation details). Does that shape work for you?

And the #1084 stderr split — thanks for the `TRYBUILD=overwrite` pointer, I'll regenerate the rest there.

schungx:
I think we're hitting on something really useful here. 

I'd propose calling the new API `new_runtime_frame` or something of that sort. Call the guard a frame guard. 

Then a new cache level will be one of the several things we may do to ensure a new isolated frame that can be rolled back automatically when out of scope. We can add other stuff to it later on. We can even have a parameter (or a builder API) to config whether to push a new caching later or not.

For example, pushing a module into `lib` can potentially be such an operation that can roll back automatically, with the original length of the lib stores inside the guard. Same with the `source`, `tag`, etc.

In other words, anything in `GlobalRuntimeState` that needs to undo can be put inside this API for one-liner simplicity. 

Something like:
```rust
#[must_use]
pub fn new_frame(&mut self, new_functions_resolution_cache: bool, rewind_modules: bool) -> RuntimeFrameGuard
```

schungx:
Update:

Come to think of it, it probably would require more hacking...

```rust
let _g = ctx.isolate_fn_resolution_cache();
let result = ctx.call_fn_raw(name, true, false, args);
```

`ctx` would be mutably borrowed by `_g` which would disallow its being used in `ctx.call_fn_raw`.

The guard would need to `Deref` to `NativeCallContext` and the code changed like this:

```rust
let ctx2 = ctx.new_isolated_frame(true, true);
let result = ctx2.call_fn_raw(name, true, false, args);
// ctx2 is dropped, ctx is restored
```

or more concisely:

```rust
let result = ctx.new_isolated_frame(true, true)
                .call_fn_raw(name, true, false, args);
```

schungx:
Based on your idea, I have put together this PR: https://github.com/rhaiscript/rhai/pull/1085

Essentially, your code would now become:

```rust
let result = ctx.new_frame().with_new_caching_layer().call_fn_raw(name, true, false, args);
```

Can you please test if this solves your problem?

schungx:
@yuvalrakavy   I wonder if you have a chance to test this out?

I have modified your test to this one-liner:

```rust
ctx.new_frame().with_new_caching_layer().with_module(module).call_fn_raw(name, true, false, args).map(Some)
```

and it passed!

schungx:
@yuvalrakavy   I have added your test to CI with the new API and it seems to work fine.  So I have merged the new API, and so I'd close this PR.

PR #1084, opened 2026-04-23
chore(ci): refresh codegen UI-test stderr for current stable rustc

yuvalrakavy:
Refresh the trybuild stderr snapshot for `codegen/ui_tests/rhai_mod_unknown_type_return` to match current stable rustc's rendering of the "builtin type with similar name" suggestion (the multi-line help block is now collapsed into a single `help: …` suffix).

The UI test failure on `upstream/main` pre-dates #1083 — verified on `c3202b62`, before the cache-isolation feature commit. Split out of #1083 per @schungx's request so the feature PR is feature-only.

No behaviour change; snapshot update only.

### DIFF
--- a/codegen/ui_tests/rhai_mod_unknown_type_return.stderr
+++ b/codegen/ui_tests/rhai_mod_unknown_type_return.stderr
@@ -2,10 +2,4 @@ error[E0425]: cannot find type `boool` in this scope
   --> ui_tests/rhai_mod_unknown_type_return.rs:12:37
    |
 12 |     pub fn test_fn(input: Point) -> boool {
-   |                                     ^^^^^
-   |
-help: a builtin type with a similar name exists
-   |
-12 -     pub fn test_fn(input: Point) -> boool {
-12 +     pub fn test_fn(input: Point) -> bool {
-   |
+   |                                     ^^^^^ help: a builtin type with a similar name exists: `bool`


schungx:
You can run with `TRYBUILD=overwrite` to update all files automatically.

yuvalrakavy:
Closing — turned out to be a wrong-direction fix on my end.

Running `TRYBUILD=overwrite cargo test -p rhai_codegen --features metadata` locally (matching CI's job config) produces zero diff against the repo snapshots, and all 64 UI tests pass. But CI still shows 57 of 64 as "mismatch" — and looking at the specific failures (e.g. `rhai_mod_unknown_type_return.rs` here shows CI's `ACTUAL` in the old multi-line help format, while the snapshot I pushed is in the new single-line format; `rhai_mod_path_attr.rs` shows CI producing one error where the snapshot expects two), it's pretty clearly a rustc-version skew between GitHub Actions' stable and my local stable. My local rustc produces newer diagnostic formatting; CI's produces older.

That means the one file I refreshed in this PR was the right direction for my local but the wrong direction for CI — I effectively flipped that one test from green-on-CI to red-on-CI. The other ~56 drifted tests were already like that before this PR (my `TRYBUILD=overwrite` would have regenerated them if my rustc disagreed with the snapshots, but it doesn't — they're calibrated to something between my local and CI).

I don't have a reliable way to regenerate the snapshots to match CI's exact rustc from here, so closing rather than keeping a yellow PR around. Happy to retry if you can share which stable you'd like pinned for the Codegen Build matrix, or regenerate on a matching env yourself.

Thanks for the `TRYBUILD=overwrite` pointer regardless — confirmed the local-vs-CI mismatch quickly.

schungx:
No worries. I'll get to it in a bit. 

It happens occasionally when Rust compiler version changes. 

This is not due to your PR.

PR #1086, opened 2026-04-30
fix: array.index_of(string) misdispatches to deprecated fn-name overload

### LINKED ISSUE #795: array.index_of: string argument got mistaken as function?
```
let array = ["hello"];
array.index_of("hello");
```

```
EXCEPTION: Function not found: hello (&str | ImmutableString | String, i64)
in call to function 'index_of' (line 2, position 7)
```

yinho999:
This PR added a value-comparison fallback for `array.index_of` when given a string argument and no script function of that name is registered.

Fixes #795

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Bug fixes
 
 * `Engine::compact_script` now properly compacts scripts with custom syntax that uses `$raw$` (thanks [`@yuvalrakavy`](https://github.com/yuvalrakavy) [`#1079`](https://github.com/rhaiscript/rhai/pull/1079)).
 * The string methods `split`, `split_rev` and their variants are now marked pure so they can be called on `const` strings ([`#1081`](https://github.com/rhaiscript/rhai/issues/1081)).
+* `array.index_of` now falls back to value comparison for string argument when no script function of that name is registered ([`#795`](https://github.com/rhaiscript/rhai/issues/795)).
 
 New features
 ------------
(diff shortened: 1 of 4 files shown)


schungx:
Suggest you put the inner function inside the `array_modules` module. May need to mark it with `#[rhai_fn(skip)]`.

Other than this it is OK!

schungx:
Some CI failing probably due to missing feature gates. 

Don't worry about the codegen tests. I can fix them later.

yinho999:
> Suggest you put the inner function inside the `array_modules` module. May need to mark it with `#[rhai_fn(skip)]`.
> 
> Other than this it is OK!


I tried to place the `index_of_start_inner` within the `pub mod array_functions` with `#[rhai_fn(skip)]`. However the codegen macro rejects every borrowed `ctx`.
- `ctx: &NativeCallContext` → *"references from Rhai in this position must be mutable"*
- `ctx: &mut NativeCallContext` → *"function parameters other than the first one cannot be passed by reference"* (`array:&mut Array` is already the first reference parameter)
- `ctx: NativeCallContext` (owned) compiles, but moves `ctx` into the helper, so the deprecated wrappers can no longer use it

Please let me know if there is another way to place it . I will be happy to work on it.

schungx:
Ah OK. So I see the prob. Leave it there then. 

But I suppose we should put in a doc comment stating that it is here temporarily to support the deprecated functions and can be removed in the next major release.

schungx:
@yinho999   A few CI tests are still failing...  They would require `#[cfg(not(feature = "no_object"))]]` and `#[cfg(not(feature = "no_function"))]]`.

Or change it to use function call syntax instead of method syntax:

```rust
 index_of([1,2,3,2], "is_two")
```

PR #1088, opened 2026-05-08
feat: support `$$` escape in string interpolation

### LINKED ISSUE #694: Escape string interpolation
Writing Bash scripts in Rhai is somewhat uncomfortable when it comes to Bash's `${VAR}` syntax:

```bash
cmake .. ${cmake_args[@]}
ninja
```

Currently I have to write

```
let script = `
  cmake .. ${"${cmake_args[@]}"}
  ninja
`;
```

Could there be some forms of escaping this syntax, either double dollar sign (`$$`) similar to writing back-ticks (``` `` ```), or a more general escape pattern that includes `\$` (or more broadly, `` \` ``)?

yinho999:
This PR added `$$` as an escape sequence inside backtick-string interpolation that emits a literal `$`.
For example, `$${...}` produces a literal `${...}` instead of triggering interpolation.

Fixes https://github.com/rhaiscript/rhai/issues/694


schungx:
I would rather not make `$$` special...  instead, I've put together another version https://github.com/rhaiscript/rhai/pull/1089 that uses `\${...}` to escape the `${...}`.

This is consistent with JavaScript.

Standalone `\$` does not do anything and still yields `\$`.  Therefore it requires three characters in a row `\${` to trigger, and `unget` one if the third one is not `{`.

I have moved your tests to the new version as well.

PR #1107, opened 2026-08-05
Fix global constants in nested call_fn functions

jaredly:
I was running into some confusing 'variable undefined' errors, and it turned out that top-level constants weren't being preserved across function calls within a module.

## Work summary

Preserve top-level constants when call_fn evaluates an AST and make them available in fresh nested script scopes. Merge encapsulated constants with active runtime constants so helper calls retain the caller environment. Add a regression test.


schungx:
This is a bit odd.  `global.constants` is supposed to be used for global constants accessed via the `global::VALUE` syntax.

But since your test does not use `global::`, so it shouldn't hit that mechanism.

In fact, after `run_ast_with_scope` shouldn't your scope contain the constant `VALUE`?  That can be used in subsequent runs of `call_fn`...

But of course this only works for the _first_ level of call.  You probably want those constants in subsequent call levels.

In that case, I'd suggest you use `global::VALUE` as the syntax is intended for exactly this scenario.

schungx:
Closing this for now as `wontfix`.  Prefer to stick with current `global::VALUE` syntax to access global constants.

jaredly:
Would you be open to a more informative error when people use this incorrectly? The current "VARIABLE is not found" is quite confusing. (edited w/ corrected message)

schungx:
Currently the error should be something like variable not found... 

Unbound is when you're trying to use `this` but the function is not called as a method.

PR #1108, opened 2026-08-07
Expose more internal items and allow unvalidated `Fn` creation 

ImTheSquid:
I am working on a stack-based VM for Rhai called [`rhaigrain`](https://github.com/ImTheSquid/rhaigrain) that needs access to some of the internals of Rhai that aren't currently exposed. rhaigrain is already working on a fork of Rhai, so I'd like to upstream. This PR adds some more exports under the `internals` feature and allows for unvalidated `FnPtr` creation. Adding these under `internals` reduces the maintenance burden, since `internals` functions have no compatibility guarantees (as referenced in the book).

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -9,7 +9,7 @@ use crate::func::native::{
 use crate::packages::{Package, StandardPackage};
 use crate::tokenizer::Token;
 use crate::types::StringsInterner;
-use crate::{Dynamic, Identifier, ImmutableString, Locked, SharedModule};
+use crate::{expose_under_internals, Dynamic, Identifier, ImmutableString, Locked, SharedModule};
 #[cfg(feature = "no_std")]
 use std::prelude::v1::*;
 use std::{collections::BTreeSet, fmt, num::NonZeroU8};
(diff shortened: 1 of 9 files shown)


schungx:
Your repo is really interesting. When you're done, would you be interested in merging it into Rhai itself? 

Or put it under the Rhai org? 

On exposing internal functions and types, I'm not particularly concerned. Let me go thru your changes in a bit.

schungx:
By the way, I think there is an internal `FnPtr` constructor called `new_unchecked` or something that allows you to bypass the checks.

schungx:
Also, would encourage you to join the Rhai Discord server so we can discuss the VM there. 

I can open up a VM section.

schungx:
Hi, I have just reviewed your proposed changes.

I am really not very comfortable with opening up so many pure implementation details, even under `internals`.

I wouldn't mind changing them to `pub(crate)`, but would hesitate if users start using them... since changing anything would break a lot of code, even though these are marked volatile.

It almost looks like the VM belongs inside the Rhai repo due to its tight integration.  What would you say we actually move it inside Rhai under a dev branch instead?

This way, the VM can take advantage of actual Rhai feature flags (like `unchecked`) for further optimizations such as turning off safety checks.

Also, eventually if the VM can be made async (which is possible for bytecodes), then Rhai would need to start registering async functions.  In other words, this will happen sooner or later, so probably sooner the better?

ImTheSquid:
Yes I'm 100% on board with merging this into Rhai, let's talk in the Discord!

PR #1109, opened 2026-08-07
Bytecode VM

ImTheSquid:
This PR merges my experiment [`rhaigrain`](https://github.com/ImTheSquid/rhaigrain) into the main Rhai repo under the `grain` module and feature flag, requested as an alternative to #1108. It will be marked as experimental. This PR also updates the Rust edition to 2021 from 2018 to support some operations that were previously forbidden by the borrow checker (such as split mutable borrows). The MSRV was not changed.


schungx:
There are some failed CI tests...  Can you look into them? 

All in all, surprisingly few global impact for such a large feature! The changes are all localized and we'll packaged.

src/module/mod.rs
```diff
@@ -373,10 +373,10 @@ impl FuncRegistration {
     ///
     /// # Parameter Examples
     ///
-    /// `"foo: &str"`   <- parameter name = `foo`, type = `&str`  
-    /// `"bar"`         <- parameter name = `bar`, type unknown  
-    /// `"_: i64"`      <- parameter name unknown, type = `i64`  
-    /// `"MyType"`      <- parameter name unknown, type = `MyType`
```
schungx, line 379: I think the two spaces at the end of the lines indicate Markdown to put a newline there. Otherwise in Vscode popups all the lines are merged into a single paragraph. 

The alternative is simply to put blank lines between each pair of lines.

src/grain/bytecode/op.rs
```diff
@@ -0,0 +1,535 @@
...
+/// place in the source, and the position for that comes from the program's
+/// [`Positions`](crate::grain::bytecode::Positions) table, keyed on the instruction's
+/// own address. Keeping it out means the diagnostics can be stripped from an
+/// artifact without touching the code.
+///
+/// Anything too wide for an operand is a `u32` index into one of the program's
+/// pools, which is also what keeps a repeated operator or name from being
+/// stored twice.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Op {
```
schungx, line 93: This is what I mean by very large VM instructions. This enum is probably 128bits or 16 bytes. 

Might need to think about squeezing it down...

ImTheSquid: This is just the program representation, if you look at the second block in the comment, this isn't the actual executed form. The real widths are at https://github.com/rhaiscript/rhai/pull/1109/changes/BASE..4ec3b7e7bb556db0c9a30655812591c41595b183#diff-3ada7e14e9756a2cc2192410b8b493536f3304cacac402dc87eb9ef74fed5849R157

Cargo.toml
```diff
@@ -147,6 +149,21 @@ required-features = ["serde"]
 name = "definitions"
 required-features = ["metadata", "internals"]
 
+# The grain harnesses, as one binary. `tests/mod.rs` wraps them in an inline
+# `mod grain`, which is what puts their module paths inside `tests/grain/`.
+[[test]]
+name = "mod"
```
schungx: Shouldn't this test be called `grain_mod` or something?

Cargo.toml
```diff
@@ -147,6 +149,21 @@ required-features = ["serde"]
...
+path = "tests/mod.rs"
+required-features = ["grain"]
+
+# Its own binary, and not negotiable: it installs a counting global allocator
+# and reads process-global counters as deltas around each call, so any test
+# allocating on another thread corrupts every number it reports.
+[[test]]
+name = "grain_allocation"
```
schungx: Not very clear what this test is...  Does it detect memory leaks?

ImTheSquid: This is just an efficiency benchmark that ensures it allocates less than the AST engine. I will rename it to `grain_allocation_efficency`

schungx:
Final thoughts:  Do you suggest creating a new branch for it?  Or just merge into `main`?

Actually merging into `main` shouldn't be a problem as the changes are extremely localized (other than the fuzzing part).  But of course a separate branch would allow for testing new ideas with more extensive changes, such as to compile dynamically-loaded script modules.  I suspect for that to happen, there'll be more changes in the core types in Rhai to accommodate bytecode script definitions in modules, or to even load modules which are pre-compiled into bytecodes.

And, as I mentioned in Discord, should we call the feature `bytecodes` instead of `grain`?  The benefit of using `grain` is that we can add new VM engines or bytecode formats later on.  The downside is the user must be educated on what `grain` is.

ImTheSquid:
I think pushing to `main` is fine, it's gated behind a feature flag and doesn't have an impact on the rest of the code. We can always improve it in a branch when we inevitably find ways to do so, and I agree that deeper changes may be needed anyways to get really big features like the ones you mentioned. I think keeping the name as `grain` makes sense since this is an enclosed feature (bytecodes, vm, compiler) coming from `rhaigrain`, whereas `bytecodes` kinda buries the lede on what is actually contained within the module. All it takes is looking at the module docs to know exactly what it is.

schungx:
Sorry, I just edited my last comment but you might have already read it.

Basically two questions:

- `grain` or `bytecodes` (or `vm`) as feature name (and src tree root name)

- new branch or merge into `main`

ImTheSquid:
I think `grain`, also because internally it has mods both named `bytecode` and `vm` so it would be a bit clunky with those names. I also think merging to main is fine in this case, nothing changes for users who don't care to use it.

schungx:
CI caught a bug in 32-bit integers mode...

schungx:
A question:  right now, how does the user use this thing?

Something like the following?

```rust
let code = engine.compile_into_grain(&ast)?;
let result = engine.eval_grain_with_scope(...)?;
```

ImTheSquid:
Example now in the `grain` module header

src/grain/mod.rs
```diff
@@ -23,6 +24,63 @@
...
+//! ```
+//! use rhai::grain::{Compiler, Vm};
+//! use rhai::{Engine, Scope};
+//!
+//! let engine = Engine::new();
+//! let ast = engine.compile("let total = 0; for i in 0..10 { total += i; } total")?;
+//!
+//! let program = Compiler::new().compile(&ast);
+//!
+//! // The `Scope` is the caller's: locals a script declares are left in it,
```
schungx: There is a typo here...  the semicolon should be after `locals`

schungx:
OK, one final round of CI...

ImTheSquid:
wait need to make sure it compiles under `no_std`, i think there are a few lingering items

schungx:
Right now, I see that your intended usage scenario is to compile a single AST into bytecodes.  Then run it on a VM separately without the source.

We need to address the scenario of loaded modules.  I'm not sure how you're handling the `import` statements and namespace-qualified calls in the VM...  I suppose they got delegated back to the engine?

In that case, we'd want to modify the `FileModuleResolver` with the ability to load stored bytecodes.  That way, the VM should probably run those functions directly once they're resolved from the resolver.

Which also means that the standard `Module` type should now support registering functions that are bytecodes.  What that means is extending `RhaiFunc` with a new variant `Grain` etc. and the ability to call it directly.

In that case, standard Rhai scripts can also load/call grain functions in a seamless manner.

Also, `rhai-run` probably should support running grain files by default.  We probably should choose a name for the grain output files?  Perhaps `.rhaigrain` or `.rhaix` or something...

In addition, perhaps a new command line tool `rhai-grain` to compile individual `.rhai` files and/or run it.

ImTheSquid:
The current convention is `.rgrn`. I agree a new tool would be nice. In terms of module resolution, that can be a later feature once `grain` gets some usage and we learn more about potential bugs or suggested use cases. Right now it just fails to compile and falls back to the regular engine, so multi-module cross-compilation support is gonna be a more complicated addition that should probably be a separate PR.

I want to get this out in experimental to library users who are interested then add the more complex stuff in the future.

schungx:
> The current convention is .rgrn

I'd suggest beginning it with `.rhai` as it would enable users to search for both Rhai scripts and compiled objects in one `.rhai*`

Alternatively, many scripting systems name their compiled files by attaching a `c` at the end so `.rhaic`... looks weird though. 

If we settle on a standard extension, we can write a standard header to the file to contain the bytecodes format and/or version (we'd want backwards compatibility). This way, Rhai can load the correct VM based on the header info.

Or we can base64 encode and make it text based, so we can still call it `.rhai`...

ImTheSquid:
Yeah the header already has versioning information so we could probably just look for the magic string at the beginning of the file. I'm pretty open to whatever since file extensions don't really matter.

schungx:
> Yeah the header already has versioning information so we could probably just look for the magic string at the beginning of the file.

Why don't we use the first 4 bytes in the file as a 4CC to identify the bytecode type...  Like `!GRN`.  So grain files are all marked as `!GRN`.

ImTheSquid:
Way ahead of you lol https://github.com/rhaiscript/rhai/pull/1109/changes#diff-df724370e809f2aaeb7c3d69f65f0585a2514e65309026d49f4d1362fe3aae12R58

schungx:
Ya I see that `RGRN`... 

Ok, I'll merge this in about 12 hours. Give some time to last minute changes, if any. 

When I get back to a desk, I'll send you an invite as an org member and give you direct commit rights. I don't think I can do it on GitHub mobile app...

ImTheSquid:
Awesome, sounds good!

schungx:
Ok you should now have commit access to `rhai`.

ImTheSquid:
Yup! I'll let you merge this PR unless you'd like me to do so.

schungx:
You can do the honors.:-D

schungx:
A few suggestions:

- Make the API `Vm::eval_with_scope` and `Vm::run_with_scope` to mirror Rhai's API. Perhaps also shortcuts without scope. 

- Not sure if the `eval_file` variants would be useful... If we consider bytecodes advanced usage then probably no need. But if we intend users to be given compiled files and then simply run them then probably yes.

- Add `Vm::call_fn` that mirrors Rhai's `call_fn` API so the user can selectively call a function instead of always having to run the whole script. Potentially also use `CallFnOptions` where make sense. 

- Can we benchmark it again?

schungx:
Also, forgot to ask. 

How do we handle `eval`? Or not at all? 

I suspect not at all right now... As it requires a parser and variable offsets will be all wracked. 

Beware that custom syntax may also modify the scope file so variables offset may be wracked anyway. So the VM needs to keep a flag on this.

ImTheSquid:
Right now `eval` dies like `import` does because it would essentially require JIT. Custom syntax is also rejected by the compiler, could also probably be fixed with JIT but that's a later problem. I've added a bench example and the functions you've suggested.

schungx:
> Right now `eval` dies like `import` does because it would essentially require JIT. Custom syntax is also rejected by the compiler, could also probably be fixed with JIT but that's a later problem. I've added a bench example and the functions you've suggested.

Yeah, I see it in the code.  `eval` and custom syntax both fails to compile to bytecodes, which should be fine for the majority of users.

However, a custom syntax definition can declare itself to not touch the scope, so perhaps it should be allowed for those cases.

And I see `call_fn` and `call_fn_with_options` are all available!  Great job!

Pls do the honors of merging!

schungx:
Actually if we get `import` working, I have a few applications that I can test the VM.  It is a bit of a limitation not to be able to import modules...

As a normal user, I would expect the VM to import modules that must also be compiled, from `.rgrn` files.

schungx:
@ImTheSquid   I just found out that the `unchecked` and `no_position` features would conflict with `grain`.  There are a bunch of methods not available when `unchecked` or `no_position` is set.

Also, I ran the bench (without `unchecked` and `no_position` which I might turn on for benchmarking):

```txt
                            walker          vm   speedup   floor   spread walker-slow  fragments
tight integer loop         106.0ms      54.4ms     1.95x   1.30x      29%     194.0ms          0
float arithmetic           537.8ms     272.4ms     1.97x   1.10x      83%     405.9ms          0
script fn calls             90.8ms      51.2ms     1.77x   1.55x      40%     106.0ms          0
switch, 4 arms             145.4ms      83.7ms     1.74x   1.40x      22%     261.3ms          0
switch, 16 arms            136.3ms      78.1ms     1.75x   1.35x      13%     283.3ms          0
branch heavy               197.8ms     107.9ms     1.83x   1.45x      12%     414.1ms          0
native callbacks            13.5ms      62.3ms     0.22x   0.25x       9%      19.8ms          0
```

So it seems roughly 2x speedup can be expected!

schungx:
I see in the code that `Vm::call_fn_with_options` is not supported for binding the `this` pointer.

That would be a pretty severe limitation...  I personally use it quite a lot, and it is very common for `call_fn` to be used in calling event drivers, all of which are written as functions with the context bound to `this`.

Is there a particular reason why it cannot be supported?

ImTheSquid:
That was just me not wanting to deal with how to deal with `this` in stack frames, I'll see how easy it could be to do. I'll also try to fix `unchecked` and `no_position`, then merge. I think `import` is a pretty big job so I'm gonna end scope here and do `import`, custom syntax, and `eval` in a separate PR.

ImTheSquid:
Lots of stuff here regarding making `this` work properly, but should be all good now. Also fixed the the feature flag compilation.

schungx:
I see the six new opcodes needed to make `this` work. Is this due to the fact that `this = v` will failed before evaluating `v` if `this` is unbound? 

Rhai did it this way to fail early and save some cycles. However we can consider this an implementation detail and Rhai can probably be changed to make it easier on the VM. This, the VM can evaluate `v` before finding out `this` does not have a target. That may remove a lot of the complications... 

Also, not sure why it is so, but I've tried declaring functions that use `this` and using it with the VM, which worked perfectly. So the VM must be able to handle function methods before.

ImTheSquid:
That's the reason for the `REQUIRE_THIS` opcode, the others are just used in the process of dealing with the semantics of `this` in the context of the stack environment where it kinda needs to be passed around like a hot potato lol. It's only a single opcode and a simple check so it's probably fine, but maybe it is unnecessary. It really depends though because if the function is native/impure then that could change a lot. The VM had some bad behavior that would allow some instances of `this` through and block others, now it consistently works with all of them.

schungx:
One more thing. I see you lifted the syntactic fn names out as an independent function and call that in each of the entry points. 

Beware that some of these entry points cross call each other, so we may be checking for syntactic function names more than once in a single call. 

That probably would not affect performance but it would still be wise to avoid it as function calls are done a lot in Rhai... All operators are function calls.

ImTheSquid:
Yeah that was just out of caution, it can be optimized later if necessary but before stuff like `type_of` would fail in very weird ways due to how dispatching works.

src/eval/expr.rs
```diff
@@ -55,6 +55,54 @@ impl Engine {
     ) -> RhaiResultOf<Target<'s>> {
         // Make sure that the pointer indirection is taken only when absolutely necessary.
 
+        // A bare script-function name is a function pointer, not a variable read.
```
schungx, line 58: Actually I don't see why this needs to be lifted out, because the effect is the same.  The next statement is a `match` which will fall into the correct case anyway.

schungx:
Ah, too late to raise the last issue.  No worries, we can handle it later.

PR #1111, opened 2026-08-09
Ensure `grain` builds in all valid feature configurations

ImTheSquid:


.github/workflows/build.yml
```diff
@@ -47,34 +47,35 @@ jobs:
...
-          - "--features testing-environ,no_custom_syntax,serde,metadata,internals,debugging"
+          - "--features testing-environ,decimal,serde,metadata,internals,debugging,grain"
+          - "--features testing-environ,unchecked,serde,metadata,internals,debugging,grain"
+          - "--features testing-environ,sync,serde,metadata,internals,debugging,grain"
+          - "--features testing-environ,no_position,serde,metadata,internals,debugging,grain"
+          - "--features testing-environ,no_optimize,serde,metadata,internals,debugging,grain"
+          - "--features testing-environ,no_float,serde,metadata,internals,debugging,grain"
+          - "--features testing-environ,f32_float,serde,metadata,internals,debugging,grain"
+          - "--features testing-environ,no_custom_syntax,serde,metadata,internals,debugging,grain"
           - "--tests --features testing-environ,only_i32,serde,metadata,internals,debugging"
```
schungx: This line can be deleted?

.github/workflows/build.yml
```diff
@@ -47,34 +47,35 @@ jobs:
         flags:
           - ""
           - "--features grain"
```
schungx: This can be deleted? Since testing-environment actually does nothing.

schungx:
The folded closures issue is a great find. It is not intended and can be considered a Rhai bug. 

The shortcut was put there as an optimization to avoid having to resolve the function every time (because the function is already known... there is only one possibility). Since simple closures are used a lot, this helps with performance.

The fact that optimization happens afterwards and essentially split the body into two - one optimized and one not - is an unintended consequence. It is definitely a bug.

And also, a closure function is not _supposed_ to be rewritten (or even rewritable) so the VM can safely ignore the cached version.

We might wrap the body under interior mutability in order to overwrite it...

Actually it doesn't really affect performance all that much because subsequent calls are cached anyway... So I guess this _feature_ can probably be removed, simplifying everything.

schungx:
https://github.com/rhaiscript/rhai/blob/main/src%2Foptimizer.rs#L1514

This is where the closure body is cloned when there is an outstanding cached reference.

In fact, removing this feature probably won't matter performance, as the function is almost always found in the same AST (and not in any other modules).

ImTheSquid:
Lets turn this into another PR, want to keep this PR focused.

PR #1113, opened 2026-08-09
Stop unnecessarily deep cloning chain root

ImTheSquid:
Improves performance to parity on this script:
```rhai
const SIZE = 10_000;

let pm = [];
pm.pad(SIZE + 1, true);

pm[0] = false;
pm[1] = false;

let tt = 0;

for p in 2..=SIZE {
    if !pm[p] { continue; }

    tt += 1;

    for i in range(2 * p, SIZE + 1, p) {
        pm[i] = false;
    }
}

tt
```

### DIFF
--- a/examples/grain_bench.rs
+++ b/examples/grain_bench.rs
@@ -120,13 +120,17 @@ const CASES: &[Case] = &[
     // of the VM, through rhai's dispatch and back into a second `Vm` with an
     // empty resolution cache — where the walker stays inside itself and reaches
     // the closure body directly. 1000 crossings per iteration.
+    //
+    // Also the only case here that indexes: the `a.push(i)` loop is a chain
+    // rooted at a local, so it is what says the root is still being walked
+    // where it lives rather than copied out and put back.
     Case {
         name: "native callbacks",
         source: "let a = []; let i = 0; while i < 500 { a.push(i); i += 1; } \
                  let b = a.map(|x| x * 2); b.filter(|x| x % 3 == 0).len()",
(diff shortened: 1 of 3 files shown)


examples/grain_bench.rs
```diff
@@ -120,13 +120,17 @@ const CASES: &[Case] = &[
     // of the VM, through rhai's dispatch and back into a second `Vm` with an
     // empty resolution cache — where the walker stays inside itself and reaches
     // the closure body directly. 1000 crossings per iteration.
+    //
+    // Also the only case here that indexes: the `a.push(i)` loop is a chain
+    // rooted at a local, so it is what says the root is still being walked
+    // where it lives rather than copied out and put back.
     Case {
```
schungx, line 127: So does this case now run at least as fast as the AST walker? 

We'd also need to do something wrt slow rust callbacks... Rhai scripts very liberally call Rust functions.

schungx:
Can't say I follow all of it but it looks the right direction, passing along an enum indicating whether it is a scope value or a detached temporary. 

Kudos! One mystery solved.

examples/grain_bench.rs
ImTheSquid: Yes it's now on par to a tiny bit faster. We can work on callback speed in another PR

ImTheSquid:
It essentially was because the root of a chain was always cloned out then re-inserted at the next part of the chain, instead of being borrowed when possible. This led to an insane number of allocations and deallocations instead of using what already existed.

PR #1120, opened 2026-08-11
VM: Speed up Rust native function calls by using FnArgsVec (smallvec)

### LINKED ISSUE #1116: VM: Speed up function calls in Rhai Grain
Right now, Rhai Grain does not obtain any speed-up when making function calls.

I find that this is probably so because `call_stacked` uses `EvalContext::call_fn_raw` to make the function call, which would actually recalculate the call hash, every single time.

The AST already contains such hashes in `FnCallExpr` and `FnCallHashes`.  These hashes are just about as stable as the hash used in `switch` statements.  When loading from disk, Rhai Grain already makes sure that hashes are stable anyway, so there is no reason not to use these hashes.

I propose adding a `Vec<FnCallHashes>` into `Program` such that these hashes don't have to be recalculated for each function call (which can be many).

Rhai Grain can use `EvalContext::_call_fn_raw` directly, which I can expose for use, as well as add a hashes parameter to reuse pre-calculated hashes.  This way, it would probably save quite a few cycles, as hashing is not cheap.

schungx:
## Use `FnArgsVec` to avoid allocations

This PR reduces `Vec` allocations during Rust native function calls by using `FnArgsVec` (which is a typedef to `smallvec`) instead to hold arguments.  This generates a 50% speed-up for function calls.

### DIFF
--- a/examples/grain_bench.rs
+++ b/examples/grain_bench.rs
@@ -83,7 +83,14 @@ const CASES: &[Case] = &[
         callbacks: false,
         floor: 1.55,
     },
-    // The VM scans its case hashes; rhai probes a hash map. Two sizes,
+    Case {
+        name: "recursive fibonacci",
+        source: "fn fib(n) { if n < 2 { n } else { fib(n-1) + fib(n-2) }} fib(28)",
+        iterations: 1,
+        callbacks: false,
+        floor: 1.50,
+    },
+    // The VM scans its case hashes; Rhai probes a hash map. Two sizes,
     // because which of those wins is a question about how many arms there
     // are, and a `switch` nobody would write is the only place the scan can
     // lose.
@@ -116,6 +123,13 @@ const CASES: &[Case] = &[
         callbacks: false,
         floor: 1.45,
     },
+    Case {
+        name: "native function calls",
+        source: "let a = 42; for i in 0..20000 { a = abs(abs(abs(abs(a)))); } a",
+        iterations: 5,
+        callbacks: true,
+        floor: 1.50,
+    },
     // The one case the VM is expected to lose. Every element is a boundary out
     // of the VM, through rhai's dispatch and back into a second `Vm` with an
     // empty resolution cache — where the walker stays inside itself and reaches
@@ -131,6 +145,35 @@ const CASES: &[Case] = &[
         iterations: 20,
         callbacks: true,
         floor: 0.64,
+    },
+    Case {
+        name: "primes",
+        source: r#"
+            const SIZE = 1_000_000;
+
+            let prime_mask = [];
+            prime_mask.pad(SIZE + 1, true);
+
+            prime_mask[0] = false;
+            prime_mask[1] = false;
+
+            let total_primes_found = 0;
+
+            for p in 2..=SIZE {
+                if !prime_mask[p] { continue; }
+
+                total_primes_found += 1;
+
+                for i in range(2 * p, SIZE + 1, p) {
+                    prime_mask[i] = false;
+                }
+            }
+
+            total_primes_found
+        "#,
+        iterations: 1,
+        callbacks: false,
+        floor: 1.55,
     },
 ];
 
--- a/src/grain/compile/mod.rs
+++ b/src/grain/compile/mod.rs
@@ -1813,16 +1813,24 @@ impl Lowering {
     /// captures the enclosing scope is closure construction, and a qualified
     /// name resolves against imported modules; neither is a plain call.
     fn is_lowerable_call(&self, call: &FnCallExpr) -> bool {
+        use crate::engine::{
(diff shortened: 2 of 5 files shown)


ImTheSquid (changes requested):
~~I think we should maybe change this to be a grain subfeature `grain-unchecked` because malicious bytecode could panic the system here. We can do it at compile time to get rid of any burden of conditionals.~~ See below

ImTheSquid:
This crashes:
```rust
    /// PoC: a map template that is not a map. The verifier proves the operand
    /// is *there*, never that it is a `Map`, and the constant pool comes off
    /// the artifact.
    #[test]
    #[cfg(not(feature = "no_object"))]
    fn poc_make_map_template_type() {
        let program = one(
            &[Op::Const(0), Op::MakeMap(0), Op::Return],
            vec![Dynamic::from(42 as INT)],
        );
        assert!(program.verify().is_ok(), "the verifier accepts this chunk");
        let err = *call(&program, None).unwrap_err();
        println!("MAKE_MAP result: {err:?}");
    }

    /// PoC: a size check that claims to be element 1 of a literal that was
    /// never started. `State` tracks operands, iterators and handlers — there
    /// is no fourth counter for the size stack.
    #[test]
    #[cfg(not(feature = "unchecked"))]
    #[cfg(not(all(feature = "no_index", feature = "no_object")))]
    fn poc_check_size_without_a_literal() {
        let program = one(
            &[
                Op::Const(0),
                Op::CheckSize {
                    index: 1,
                    map: false,
                },
                Op::Return,
            ],
            vec![Dynamic::from(42 as INT)],
        );
        assert!(program.verify().is_ok(), "the verifier accepts this chunk");

        // A host that sets no limit never reads the running total, so the
        // reachable case is the one that does.
        engine.set_max_array_size(100);
        let mut options = CallFnOptions::new().eval_ast(false);
        options.this_ptr = None;
        let result: Result<Dynamic, _> =
            Vm::new(&engine).call_fn_with_options(options, &mut Scope::new(), &program, "f", ());
        println!("CHECK_SIZE result: {:?}", result.map(|_| ()));
    }

    /// PoC: a chain index step naming an operand slot the chain does not have.
    /// `check_chain_indices` skips `Step::Index` on the grounds that its
    /// operands are stack offsets, and nothing else bounds them.
    #[test]
    #[cfg(not(feature = "no_index"))]
    fn poc_chain_index_operand_out_of_range() {
        use crate::grain::bytecode::Root;

        let chain = Chain {
            root: Root::Temporary,
            steps: vec![Step::Index {
                operand: 500,
                pos: rhai::Position::NONE,
                bracket: rhai::Position::NONE,
            }],
            tail: Tail::Read,
            operands: 1,
        };
        let program = program_with_chains(
            &[&[Op::Const(0), Op::Const(1), Op::Chain(0), Op::Return]],
            vec![
                Dynamic::from(0 as INT),
                Dynamic::from_array(vec![Dynamic::from(7 as INT)]),
            ],
            vec![chain],
        );
        assert!(program.verify().is_ok(), "the verifier accepts this chunk");
        let err = *call(&program, None).unwrap_err();
        println!("CHAIN result: {err:?}");
    }
}

```

For the first one restoring the check should be fine, it's per-map literal. For the second and third you can move the check into the verifier. I take back needing a new feature for this, but we need to really prove correctness of the verifier now.

schungx:
> I think we should maybe change this to be a grain subfeature `grain-unchecked` because malicious bytecode could panic the system here. We can do it at compile time to get rid of any burden of conditionals.

I think we can do this with the `unchecked` feature.  In that case we can leave all the `_unchecked` versions out and simply but gates on the version that return `Result`.  Under `unchecked`, that simply returns `Ok`, which the compiler will optimize away the error check (I hope).

On the other hand, Rhai already has a whole bunch of places where panics may happen.  The premise of Rhai has never been "no panics", but that it will not intentionally panic -- therefore any panic is a bug.

Here, I think the premise is the same: the bytecodes are verified to the best of effort, so panics should not happen.  If it happens, then it is a bug in the VM that needs to be fixed.

ImTheSquid:
Ok then let's at least fix the bugs I posted and then we can merge this, the speed gains are really good. We should probably do some more analysis on the verifier to harden it in the future.

schungx:
I have reversed the error-handling changes.  Now it is only the Vec allocations.  We can merge this first.

I agree probably we would need some more scrutiny for the error-handling stuff.  I'll open a new PR with an alternative way of doing it.

ImTheSquid (approved):

ImTheSquid:
Let's put a pin in the verifier, I'd love to keep those speed gains but they need to be tested well.

schungx:
> Let's put a pin in the verifier, I'd love to keep those speed gains but they need to be tested well.

Yup.  I think I'd do it in a different way that would keep all the error handling intact, but remove them in `unchecked` builds.

schungx:
@ImTheSquid   I think this one can be merged first.  It does not involve introducing any potential panics.

ImTheSquid:
Agreed

PR #1122, opened 2026-08-12
Reduce error handling if unchecked

schungx:
## Description

This is a draft PR to speed up certain lookups during the VM running stage when the bytecodes are known to be well-formed.

All such are gated only under the `unchecked` feature.

Essentially, some of the lookup functions (e.g. `Program::constant`) would use the original code without the `unchecked` feature, but will use more optimized direct lookup code under `unchecked` (which can panic if there is a bug in the VM).

In both scenarios, the one uncompromising premise is that malicious _user_ inputs can never cause a panic.  Therefore, panics can only happen if there is a bug in the core system code.

## Performance

Although the checks are still performed under the faster path, they panic instead of returning `None`.  Thus, returning a constant `Some` value and turning on inlining allows the compiler to remove the subsequent checks for `None` at the call site, as well as the error handling return.

Thus `program.constant(index).ok_or_else(|| malformed(format!("no constant {index}")))?` will hopefully be optimized into a single `&program.constants[index]`, which can potentially enable further follow-on optimizations.

Benchmarked this and it looks like another 10-20% speed gain on average.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,16 +11,22 @@ This is a work-in-progress and is not yet fully integrated into the main `Engine
 
 It is gated under the `grain` feature flag.
 
+Rhai Grain VM
+-------------
+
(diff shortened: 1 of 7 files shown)


schungx:
Rebased so I'll open a new PR.

PR #1125, opened 2026-08-13
VM: Fix chain assignment evaluation order

### LINKED ISSUE #1115: VM: Fuzz errors detected
From OSS-Fuzz:

# Error 1

`roundtrip` failed on input script:
```rust
S[K]=b

the VM disagrees with the walker on:
	S[K]=b
	  left: "err ErrorVariableNotFound(\"K\", 1:3) []"
	 right: "err ErrorVariableNotFound(\"b\", 1:6) []"
```

I believe this is due to the LHS or RHS being evaluated first...

# Error 2

The setter is being called with  no arguments (requires two):

Image

This seems to be the `RGRN` file auto-generated.

# Error 3

There are some fuzz errors due to stack overflow in `generated`.

It seems like `generated` can get into an infinite loop and either overflow the stack or consume all memory.

schungx:
Closes #1115.

There is a difference between how the AST interpreter and the Grain VM handle assignment to chained expressions.

For instance, when running this script:
```rust
S[K]=b;
```

The AST interpreter will come back with `variable 'b' not found` because it evaluates the value expression first.

Rhai Grain will come back with `variable 'S' not found` because it evaluates the lvalue expression first.

This PR patches the VM such that it follows the AST interpreter's order and evaluates the value expression first, stashing the result into an temporary slot, processes the lvalue chain and then finally assigning the result into the chain target.

### DIFF
--- a/src/grain/compile/mod.rs
+++ b/src/grain/compile/mod.rs
@@ -419,7 +419,7 @@ impl Lowering {
                 return false;
             }
 
-            let value_name = self.strings_interner.get("$CHAIN_SET_VALUE$");
+            let value_name = ImmutableString::from("$CHAIN_SET_VALUE$");
             let value_name_index = self.push_name(value_name.clone());
             let value_slot = self.slots.declare(value_name);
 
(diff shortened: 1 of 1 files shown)


schungx:
OK, I'll do this on a separate branch instead, so this won't affect my patching other parts of Rhai, fixing bugs found by the VM.

PR #1129, opened 2026-08-14
Greatly improve error reporting and portability

ImTheSquid:
Up until now Grain has had issues where it doesn't correctly report errors, and after stripping it was impossible to properly resymbolize across a serialization boundary. This PR adds a sidecar with a debug ID that is associated with a program. This sidecar can be serialized. In addition, stack traces are now recorded, so the full context is available for debugging after a crash.


schungx:
I'm in the process of implementing `Expr::Coalesce` (and possibly add the coalescing operators such as `?[...]` and `?.prop`.  I found out coalescing is not lowered, for some reason, and it is used a lot in scripts.

I'll rebase my PR on top of this after merging this.

schungx:
In `flatten_chain`, there is this:

```rust
if flags.intersects(ASTFlags::NEGATED) {
    return None;
}
```

Therefore, coalescing chaining such as `a?[b]?.prop` are all ignored and kept a fragment instead.  Not sure why it was originally done this way.

I found that I'd have to add a new instruction `SKIP_IF_NOT_UNIT` which does not pop the stack but merely _inspects_ it, jumping if it is not `()`.  This instruction implements `??` (`Expr::Coalesce`) nicely.

Let me see if the chain coalescing can be done similarly...

PR #1130, opened 2026-08-14
Make sure no-std builds with everything

schungx:


thalesfragoso:
Commit `9e5a89d` seems kinda of breaking, e.g. now `rhai::Map` doesn't implement `serde::Deserialize` anymore, even with the `serde` feature.

I guess I could enable the feature in `smartstring` myself to fix this, but I wanted to let you know anyways.

schungx:
Yeah, I found out myself also. Should put a note on. Thanks for the reminder. 

It is unfortunate that `smartstring/serde` does not support `no-std`, which means I cannot automatically turn it on with the `serde` feature, otherwise `no-std` won't be able to use `serde`.

The solution, as you discovered, is to manually specify `smartstring` with `serde`.

PR #1132, opened 2026-08-14
VM: add support for coalescing and capture-parent-scope

schungx:
# Coalesce Support

This PR adds support for coalescing operators, such as `??`, `?.` and `?[...]` to the VM.  Previously, coalescing operators are not lowered.

## New opcode

A new opcode, `Op::SkipIfNotUnit` is introduced.  This opcode is named `Skip` because, unlike `Jump` opcodes, it does not pop the value off the stack.  Instead, it only inspects the value and jumps if the value is not `()`.

This new opcode is necessary to implement the `??` coalescing operator.

## Version is bumped

Artifact version is bumped because a new flag is written for the chaining types (`Step::Index`, `Step::Property` and `Step::Method) which makes it non backwards compatible with previous written artifacts.

## Implementation

`??` is implemented by the new `Op::SkipIfNotUnit` opcode.

`?[...]` and `?.` are implemented by adding a new `flags` field to the chaining types and short-circuiting where the receiver is `()`.

# Capture Parent Scope

Rhai has a feature that, if you put a `!` in a function call, it acts sorta like a "macro" -- i.e. it appears to run in-place, being able to see all variables visible at the call site.

```rust
// A function that uses the undeclared variable 'offset', which normally should raise an error.
fn offset_value(x) {
    let local_var = 999;
    x + offset
}

let offset = 10;    // Declare local variable

let result = offset_value!(42);    // call with '!'
```

The result is 42 + 10 = 52 instead of an error, because Rhai takes the function call as _capturing the parent's scope_.

Essentially it is equivalent to:

```rust
let offset = 10;    // Declare local variable

let result = {
    let local_var = 999;
    42 + offset
};
```

## Opcode changes

Four new `_CAPTURE` versions of the `Op::CALL_XXX` opcodes are introduced.

## Implementation

The flag `capture_parent_scope` has always been available for `Expr::FnCall`.  When it is set to `true`, call functions with the current scope instead of creating an empty new one.


src/grain/format/abi.rs
```diff
@@ -151,17 +168,19 @@ impl Abi {
...
-            let bit = differing.trailing_zeros() as usize;
-            // A bit past the table means the writer knew a flag this build does
-            // not. Reporting it as unknown beats indexing out of bounds.
-            let flag = FLAGS
-                .get(bit)
-                .map_or("an unknown restriction", |(name, _)| *name);
+        let diff = self.features ^ host.features;
+        if !diff.is_empty() {
+            // A bit not in the table means the writer knows a flag this build does not.
+            // Reporting it as unknown.
```
ImTheSquid: I don't think we should actually have this. A more capable host could compile a script that doesn't use any feature the target doesn't have. For instance, the target may have `no_module`, but if the host doesn't compile a module, there shouldn't be an issue.

src/grain/bytecode/code.rs
```diff
@@ -37,136 +37,159 @@ use std::prelude::v1::*;
 /// Operators get their own tag rather than an optional field, so the common
 /// call pays nothing for the one that carries a token.
 pub mod tag {
-    /// [`Op::Const`](super::Op::Const).
-    pub const CONST: u8 = 0x01;
+    // 0x00-0f - Constants, load/store/declare, stack operations
```
ImTheSquid: I am unsure about reordering these. I don't think these should really have an order because otherwise we will get in the habit of reordering them when we add a related tag to a group, instead of just appending them to the end.

src/grain/bytecode/verify.rs
```diff
@@ -234,12 +234,13 @@ fn verify_chunk(
         high_water = high_water.max(depth);
 
         let op = code::decode(code, at).ok_or(VerifyError::Undecodable { at })?;
-        let (pops, pushes) = effect(&op, pools);
+        let (requires, pops, pushes) = effect(&op, pools);
+        assert!(requires >= pops);
```
ImTheSquid: Naked assert is probably bad, should probably convert to debug assert (so it doesn't get shipped in prod) then add a message.

ImTheSquid (changes requested):

src/grain/bytecode/code.rs
schungx: Ok, will revert

src/grain/format/abi.rs
schungx: This is essentially the previous behavior, only translated into bitflags instead of hand-rolled flags.

I do agree we should probably be smarter in handling these... Maybe some form of capability flags they can collect what features are being used in a script.

I suggest just leave this here for the moment. We can switch to capabilities based flags later on...

src/grain/bytecode/verify.rs
schungx: I can convert it into an error... But this really is something that should not happen as the table has hard coded figures, do technically the assert will never fail unless we screw up big time.

schungx: Converted into Can't Decode error.

src/grain/format/abi.rs
schungx: We should think of a scheme that, instead of using feature flags, we would be using capabilities flag, such that we can match those capabilities against feature flags of the running instance.

For example, `USE_ARRAYS` can be turned on if we detect any indexing or arrays being used in the script.  Then it can be paired with `no_index` to check if the artifact can be loaded.

Leave this as the previous behavior for now, and I'll add the capabilities flags in a later PR.

schungx:
@ImTheSquid   I've reversed the re-numbering of the instructions, as well as not using any `assert!`.

Based on your idea, I've started working on a "capabilities" system that identified the syntax features used by an AST.  Then it'll be written into the artifact instead of the feature flags.  A feature not used by an artifact and not built into the client still runs fine, even if the artifact was originally written by a build that enabled such a feature.

So for example, if an artifact is written by a normal build, but loaded in a build with `no_index`, it still loads fine if the artifact does not use arrays at all.

Matching capabilities is fine, but we'd run into the problem of detecting corrupted bytecode, which may corrupt the capability flags, turning a required cap into not. In that case, the verifier needs to double check instructions such that they always abide by the declared capabilities.

schungx:
@ImTheSquid   while refining this, I discovered that the `capture_parent_scope` flag for function calls is not implemented by the VM -- it became a residual fragment.

It wasn't hard to implement, so I just added that in.  Now the VM really has feature parity with the AST walker.

src/grain/format/abi.rs
ImTheSquid: I don't think we should add a half-baked feature that is subject to change in an unrelated PR. This should be its own thing.

src/grain/bytecode/code.rs
```diff
@@ -218,7 +220,7 @@ static WIDTHS: [u8; 256] = {
     widths[tag::REQUIRE_THIS as usize] = 1;
     widths[tag::ASSIGN_THIS as usize] = 1;
     widths[tag::ASSIGN_THIS_OP as usize] = 3;
-    widths[tag::CALL_THIS_REF as usize] = 4;
+    widths[tag::CALL_THIS_REF as usize] = 5;
```
ImTheSquid: Are we sure we want to make all function calls a larger width just on the off chance they use the super rare scope capture syntax? I feel like that's a bad trade.

ImTheSquid (changes requested):

src/grain/bytecode/code.rs
schungx: Good point.

But although the feature is niche, some people may be relying on it. I think it is important for the VM to achieve feature parity with the walker as much as possible.

Actually it should be impossible for `CALL_OP` to capture, so I'll back that out. That leaves four op tags.

What we can do is to add capturing versions of those op tags. That would mean four new op tags and the encoding no longer needs to change widths.

Let me do that and see.

src/grain/format/abi.rs
schungx: OK, I'm backing it out.  You're right, this should go into its own PR.

src/grain/bytecode/code.rs
schungx: OK, done.  Introduced four new instructions, the `_CAPTURE` versions of the `OP_CALL_XXX` codes.

Instruction encoding widths are no longer widened.

src/grain/bytecode/code.rs
```diff
@@ -486,17 +512,31 @@ pub fn assemble(ops: &[Op]) -> Result<(Vec<u8>, Vec<u32>), AssembleError> {
...
                 receiver,
+                capture_parent_scope,
             } => {
                 // `this` is a register and needs no operand to address it, so
                 // its encoding is the other two minus the trailing `u16`.
                 let operand = match receiver {
+                    Receiver::Local(slot) if *capture_parent_scope => {
+                        Some((tag::CALL_LOCAL_REF_CAPTURE, *slot))
+                    }
                     Receiver::Local(slot) => Some((tag::CALL_LOCAL_REF, *slot)),
```
ImTheSquid: I feel like this is kinda an antipattern. Shouldn't it be more like
```rust
Receiver::Local(slot) => Some((if *capture_parent_scope {tag::CALL_LOCAL_REF_CAPTURE} else {tag::CALL_LOCAL_REF}, *slot))
```
same for following occurrences

src/grain/vm/mod.rs
```diff
@@ -3305,22 +3394,42 @@ impl<'e> Vm<'e> {
...
+                | code::tag::CALL_THIS_REF_CAPTURE => {
                     let name_index = u32::from(small(1)?);
                     let argc = code[pc + 3] as usize;
                     // `this` is a register, so this one carries no operand for
                     // the receiver and is two bytes shorter.
-                    let receiver = match tag {
-                        code::tag::CALL_LOCAL_REF => Receiver::Local(small(4)?),
-                        code::tag::CALL_NAMED_REF => Receiver::Named(u32::from(small(4)?)),
-                        _ => Receiver::This,
+                    let (receiver, capture_parent_scope) = match tag {
```
ImTheSquid: Definitely a nit but i think the `capture_parent_scope` should just be a call to `matches!()`

ImTheSquid (changes requested):
Small style nits, looking good for the most part!

src/grain/bytecode/code.rs
schungx: I prefer listing the combinations out. It looks clearly to me, just personal style. 

I'll change it as you suggest. No worries 😁

src/grain/vm/mod.rs
schungx: You're right!

schungx:
@ImTheSquid   Changed made.  Pls check it for one final round.

ImTheSquid (approved):

PR #1134, opened 2026-08-17
Fix function currying behavior on native functions

ImTheSquid:
Rhai had a bug where if you curried a function that was passed into a native function, the curried argument would be added *after* the other arguments, not before (like the normal behavior). This PR fixes that.

Old behavior:
```rs
engine.register_fn("subtract", |a: INT, b: INT| a - b);

assert_eq!(engine.eval::<INT>(r#"Fn("subtract").curry(10).call(1)"#).unwrap(), 9);

assert_eq!(
    engine
        .eval::<rhai::Array>(r#"[1, 2, 3].map(Fn("subtract").curry(10))"#)
        .unwrap()
        .into_iter()
        .map(|v| v.as_int().unwrap())
        .collect::<Vec<_>>(),
    [-9, -8, -7]
);
```

New behavior:
```rs
engine.register_fn("subtract", |a: INT, b: INT| a - b);

assert_eq!(engine.eval::<INT>(r#"Fn("subtract").curry(10).call(1)"#).unwrap(), 9);

assert_eq!(
    engine
        .eval::<rhai::Array>(r#"[1, 2, 3].map(Fn("subtract").curry(10))"#)
        .unwrap()
        .into_iter()
        .map(|v| v.as_int().unwrap())
        .collect::<Vec<_>>(),
    [9, 8, 7]
);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Bug fixes
 ---------
 
 * Fixes bug in `Engine::compact_script` that generates invalid compacted scripts due to missing spaces between ambiguous operators (thanks [`@yuvalrakavy`](https://github.com/yuvalrakavy) [`#1106`](https://github.com/rhaiscript/rhai/pull/1106)).
+* Fixes bug where a curried function pointer passed to a native iteration function (e.g. `map`, `filter`, `reduce`) received the element _before_ the curried arguments, the reverse of what `FnPtr::call` does.
 
 Enhancements
 ------------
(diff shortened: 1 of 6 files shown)


schungx:
This will fail on this:

```rust
fn sub(b) { this - b }
[1, 2, 3].map(Fn("sub").curry(10))
```

That's because `this` for scripts (or closures) should be bound to the item, instead of calling `sub` with three arguments.

schungx:
I think, in the majority of cases, function pointers are used to point to scripted functions. Therefore if we err, we err towards making scripted functions work, which is how it is right now.

schungx:
Closing in favor of #1139.

PR #1138, opened 2026-08-18
VM: Implement missing syntactic functions

schungx:
The syntactic functions `is_def_var` and `is_def_fn` are missing from the VM.

This PR implements them.

### DIFF
--- a/src/grain/vm/mod.rs
+++ b/src/grain/vm/mod.rs
@@ -2234,6 +2234,125 @@ impl<'e> Vm<'e> {
         Ok(())
     }
 
+    // Some syntactic calls can be self-implemented or short-circuited.
+    fn call_syntactic(
+        &mut self,
+        program: &Program,
+        name: &str,
+        argc: usize,
+        first: usize,
+        scope: &mut Scope,
+        pos: Position,
+    ) -> Result<Option<Dynamic>, Box<EvalAltResult>> {
+        match name {
+            #[cfg(not(feature = "no_closure"))]
+            crate::engine::KEYWORD_IS_SHARED => {
+                return Err(EvalAltResult::ErrorFunctionNotFound(name.to_string(), pos).into())
+            }
+            crate::engine::KEYWORD_IS_DEF_VAR => {
+                if argc != 1 {
+                    return Err(EvalAltResult::ErrorFunctionNotFound(name.to_string(), pos).into());
+                }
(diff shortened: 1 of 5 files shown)


src/grain/vm/mod.rs
```diff
@@ -2349,20 +2461,23 @@ impl<'e> Vm<'e> {
...
-                // Reuse parent scope.
-                scope
-            } else {
-                // Create detached scope.
-                new_scope = Scope::new();
-                &mut new_scope
-            };
-
-            let value = self.call_stacked(program, name_index, argc, first, use_scope, pos);
+            // Check if it is a built-in syntactic function.
```
ImTheSquid: 3x duplicated code, should be extracted into a function

src/grain/vm/mod.rs
```diff
@@ -2206,6 +2206,125 @@ impl<'e> Vm<'e> {
...
+        name: &str,
+        argc: usize,
+        first: usize,
+        scope: &mut Scope,
+        pos: Position,
+    ) -> Result<Option<Dynamic>, Box<EvalAltResult>> {
+        match name {
+            #[cfg(not(feature = "no_closure"))]
+            crate::engine::KEYWORD_IS_SHARED => {
+                return Err(EvalAltResult::ErrorFunctionNotFound(name.to_string(), pos).into())
```
ImTheSquid: we keep track of shared values, why is this unimplemented?

ImTheSquid (changes requested):

src/grain/vm/mod.rs
schungx: True. This shouldn't happen. It is there as a catch-all. I'll change that to return an error instead.

src/grain/vm/mod.rs
schungx: LOL... It was the better of two evils... I originally had the syntactic code in `call_stacked`. However, then I had to duplicate the parent-scope capturing code. So I extracted it out into its own function. How I have this call in multiple places.

Let me think about how to make it more elegant...

schungx:
OK, I fixed up the two issues.  Mainly, the `call_syntactic` and `call_stacked` pair is now extracted into a single `call_syntactic_and_stacked`...  (I know, not very creative here)

Don't worry about the other small changes...  they show up due to small commits on `main`.

ImTheSquid (approved):

PR #1140, opened 2026-08-19
Clone a chain index before the getter consumes it

ImTheSquid:
`index_by_reference` cloned the index for the write-back setter after calling `get_indexed_mut`, which binds a native's by-value parameter by `take`. The setter was called with `()`, found nothing, and `call_indexer_set` swallowed the error. The write was dropped silently. Rhai clones before the same call (`eval/chaining.rs:706`). Only the paths that can write pay for it.

### DIFF
--- a/src/grain/vm/mod.rs
+++ b/src/grain/vm/mod.rs
@@ -1440,6 +1440,13 @@ impl<'e> Vm<'e> {
         let assigning = last && value.is_some();
         let mut detached = Scope::new();
 
+        // Cloned *before* the call, as Rhai clones it (`eval/chaining.rs:706`).
+        // `get_indexed_mut` may reach a custom indexer, and a native's by-value
+        // parameter is bound by `take` (`func/register.rs:69`) — so afterwards
+        // there is nothing left to address the setter with. Only the paths that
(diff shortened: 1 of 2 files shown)


schungx (approved):
LGTM!

EDIT: See below for another change to fix a related bug.

src/grain/vm/mod.rs
schungx: Change this call to:
```rust
let _ = call(self, setter, &mut [target, &mut temp]).or_else(|err| match *err {
    // Fail silently if the property is read-only, as Rhai does (`eval/chaining.rs:1039`).
    EvalAltResult::ErrorDotExpr(..) => Ok(Dynamic::UNIT),
    _ => Err(err),
})?;
```

This will also fix the problem that a missing property setter (readonly property) will silently stop a write-back (just like for indexers) instead of failing with a readonly error.

schungx (changes requested):
Fix another big in one go.

src/grain/vm/mod.rs
schungx: I had this in a PR but haven't yet pushed it.  I see you have most of my changes already (fixing the consumed index), so I suggest just add this change here.

PR #1141, opened 2026-08-19
Flatten a declaration's initializer in the VM

ImTheSquid:
`DECLARE_LOCAL` pushed the popped value as-is where Rhai flattens (`eval/stmt.rs:438`), so a native handing back an already-shared cell left the local shared. Loading a variable flattens either way, so reaching this needs a native: `shared_cell` in the corpus engine.


src/grain/vm/mod.rs
schungx: There is a warning of `program` being unused under certain features flags.  Better fix it also here, perhaps change it to `_program`?

schungx (approved):
I understand that this is fixing behavior differences between Rhai and the VM.

PR #1142, opened 2026-08-19
Improve debugging support for the VM

ImTheSquid:
Add hooks to pass the debugging tests that previously failed. Also adds a disassembly example.


schungx:
LGTM...  But need to fix the CI.

I wasn't expecting to support the debugger interface with Grain, but if it can, why not!

Kudos!

schungx (changes requested):
LGTM, but fix CI.

PR #1143, opened 2026-08-19
Resolve ambiguity in function pointer resolution.

ImTheSquid:
Grain had an issue where its use of `Normal` function pointers led to resultion ambiguity that caused incorrect behavior. For example:

```ts
let t = 0; [1,2,3].for_each(|i| t += i); t
```

Before, Grain would bind `i` to `this`, resulting in an incorrect value for `t` of 6 instead of 3. Now this no longer happens due to my introduction of `FnPtrType::Compiled`, which properly tracks function pointer information for Grain calls.

### DIFF
--- a/src/types/fn_ptr.rs
+++ b/src/types/fn_ptr.rs
@@ -26,6 +26,22 @@ pub enum FnPtrType {
     /// Pre-calculated hash of a script-defined function.
     #[cfg(not(feature = "no_function"))]
     Script { num_params: usize, hash: u64 },
+    /// A function the Grain VM compiled, addressed in its own table.
+    ///
+    /// Distinct from [`Script`][FnPtrType::Script], which says two things at
+    /// once: how many parameters the function takes, *and* that it resolves to
+    /// an AST body at a hash in `global.lib[0]`. A compiled function has the
+    /// first and not the second, so claiming `Script` would be a lie the
+    /// dispatcher acts on.
+    ///
+    /// The program travels with the pointer rather than the index alone: a
+    /// pointer escapes to the host, and a bare index into a table it no longer
+    /// has would address whatever happened to be there.
+    #[cfg(all(feature = "grain", not(feature = "no_function")))]
+    Compiled {
+        program: crate::grain::SharedProgram,
+        index: u32,
+    },
     /// Embedded native Rust function.
     #[cfg(not(feature = "sync"))]
     Native(Shared<dyn Fn(NativeCallContext, &mut FnCallArgs) -> RhaiResult + 'static>),
@@ -43,6 +59,11 @@ impl fmt::Display for FnPtrType {
             Self::Normal => f.write_str("Fn"),
             #[cfg(not(feature = "no_function"))]
             Self::Script { .. } => f.write_str("Fn*"),
+            // Its own mark rather than `Fn` or `Fn*`: the first would hide that
+            // the pointer knows its own shape, the second would claim an AST
+            // body it does not have.
+            #[cfg(all(feature = "grain", not(feature = "no_function")))]
+            Self::Compiled { .. } => f.write_str("Fn#"),
             Self::Native(..) => f.write_str("Fn"),
         }
     }
--- a/src/types/fn_ptr.rs
+++ b/src/types/fn_ptr.rs
@@ -38,11 +38,18 @@ pub enum FnPtrType {
     /// pointer escapes to the host, and a bare index into a table it no longer
     /// has would address whatever happened to be there.
     ///
-    /// What this variant is *for* is declaring the callee's shape, which only
-    /// [`FnPtr::declared_shape`] and its one caller read. Everywhere else — the
-    /// dispatch arms in `func/call.rs`, and [`FnPtr::call_raw`] — it deliberately
-    /// falls through to dispatch by name, exactly as [`Normal`][FnPtrType::Normal]
-    /// does, reaching the chunk through the wrapper the VM registered for it
+    /// Only ever an *anonymous* function. A name a script could have written is
+    /// late-bound — `Fn("f")` is whatever `f` is at the arity the caller asks
+    /// for — and declaring a shape would settle that here instead, picking a
+    /// different function from the one Rhai would have found. See
+    /// `Vm::compiled_pointer`.
+    ///
+    /// What this variant is *for* is declaring the callee's shape, and two
+    /// places read it: [`FnPtrType::declared_params`] for how many parameters the
+    /// chunk takes, and [`FnPtr::call_raw`] for whether it can be handed a
+    /// receiver. The dispatch arms in `func/call.rs` ignore it and fall through
+    /// to dispatch by name, exactly as [`Normal`][FnPtrType::Normal] does,
+    /// reaching the chunk through the wrapper the VM registered for it
     /// (`grain/vm/callback.rs`). That is not an oversight: routing a call
     /// straight into the chunk is a larger change, and leaving those sites alone
     /// keeps this variant a strict improvement on the name-only pointer rather
@@ -94,6 +101,29 @@ impl FnPtrType {
             Self::Script { num_params, hash } if *num_params == num_args => {
                 global.lib[0].get_script_fn_by_hash(*hash)
             }
+            _ => None,
+        }
+    }
+
+    /// How many parameters the callee declares.
+    ///
+    /// [`None`] where the pointer cannot say — a bare name, or a native — which
+    /// leaves the caller to guess a shape and correct itself from the resulting
+    /// `ErrorFunctionNotFound`.
+    ///
+    /// Says nothing about the receiver: whether the callee can be handed one is
+    /// [`FnPtr::call_raw`]'s to decide, and it decides the same way wherever the
+    /// call came from.
+    #[must_use]
+    fn declared_params(&self) -> Option<usize> {
+        match self {
+            #[cfg(not(feature = "no_function"))]
+            Self::Script { num_params, .. } => Some(*num_params),
+            #[cfg(all(feature = "grain", not(feature = "no_function")))]
+            Self::Compiled { program, index } => program
+                .functions()
+                .get(*index as usize)
+                .map(|f| f.params.len()),
             _ => None,
         }
     }
@@ -491,6 +521,31 @@ impl FnPtr {
             );
         }
 
+        // A compiled chunk that does not read the receiver cannot be handed one.
+        //
+        // It is reached by name, through a wrapper registered at the chunk's own
+        // arity (`grain/vm/callback.rs`), so a receiver becomes argument zero and
+        // the call is one argument too wide to resolve at all. Rhai's own pointer
+        // has nothing to decide here: a script body binds `this` whether or not
+        // it reads it.
+        //
+        // Here rather than in [`FnPtr::call_raw_with_extra_args`] because this is
+        // the entry every caller reaches — including a host's native, which can
+        // hand over a receiver without going through the argument-shaping above.
+        #[cfg(all(feature = "grain", not(feature = "no_function")))]
+        if let FnPtrType::Compiled { ref program, index } = self.typ {
+            // Read through the index rather than stored beside it: the pointer
+            // carries the program, so the answer is one bounds check away and a
+            // copy in the pointer would be a second place for it to be wrong.
+            if !program
+                .functions()
+                .get(index as usize)
+                .map_or(false, |f| f.takes_this)
+            {
+                this_ptr = None;
+            }
+        }
+
         // Embedded native Rust function?
         match self.typ {
             FnPtrType::Native(ref func) => {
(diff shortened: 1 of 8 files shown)


schungx:
Yes, I found that too.  Because VM scripted functions are actually registered as a native callback, it threw off Rhai's logic.  Therefore we have problems with functions like `reduce` and, of course, your example above.

Also it necessitates very convoluted things like `makes_fn_pointers` and `takes_this`...  Routing stuff back to the AST walker.

I have been thinking of introducing a new VM-callback type.  You beat me to it!

schungx (approved):
In this case, can we get rid of `makes_fn_pointers` and `takes_this`, thereby simplifying the VM a whole lot?

At least I believe `makes_fn_pointers` would have no need to exist, since its whole existence is due to needing to send the callback eval back to the AST walker.  Also, I would consider it quite dangerous, as functions may create FnPtr's as their return values and the VM would have no way to predict whether function pointers will ever exist inside a program.

Also, there should now be no distinction between `run` and `run_with_callback`, as we can always provide a wrapper module by default...  This would simplify the VM API...

src/types/fn_ptr.rs
```diff
@@ -582,48 +640,55 @@ impl FnPtr {
...
-                            let mut args2 = FnArgsVec::with_capacity(args.len() + 1);
-                            if move_this_ptr_to_args == 0 {
-                                args2.push(this_ptr.clone());
-                                args2.extend(args);
-                            } else {
-                                args2.extend(args);
-                                args2.insert(move_this_ptr_to_args, this_ptr.clone());
-                            }
-                            return self.call_raw(ctx, None, args2);
+        if let Some((num_params, binds_this)) = self.declared_shape() {
```
schungx: I suggest we move this mechanism into `call_raw` itself, instead of inside `_call_with_extra_args`.

That is because `call_raw` is the main work-horse, while `_call_with_extra_args` is only intended for flexible-args type of callbacks.  Most callbacks do not use those, but they'd still need to distinguish a VM-registered callback (because of where to put the `this` pointer).

Therefore most callbacks will actually land back into `call_raw` and you have the same problem that `declared_shape` is designed to fix. So move it to `call_raw` instead by removing the `this_ptr` if it is not necessary.

In fact, it can simply be like this:

```rust
// Inside `call_raw`...
// Embedded native Rust function?
match self.typ {
    FnPtrType::Native(ref func) => {
       :
    }
    #[cfg(all(feature = "grain", not(feature = "no_function")))]
    FnPtrType::Compiled { ref program, index } => {
        // Do we have to search the functions every time?
        // Can't we just store a `bind_this`  boolean here?
        if !program
            .functions()
            .get(index as usize)
            .map(|f| f.takes_this)
            .unwrap_or(false)
        {
            // The VM-compiled function has no need for `this`,
            // so remove `this_ptr`!
            this_ptr = None;
        }
    }
    _ => ()
}

// If `FnPtrType::Compiled` stores a boolean `bind_this` instead,
// it becomes this one-liner: FnPtrType::Compiled { bind_this: false } => this_ptr = None

```

`_call_with_extra_args` can revert back to previous, because all that's need to handle the situation would already be done.

There will also be no need for `declared_shape` as it can simply be inlined into the change as above.  So the changes to `fn_ptr.rs` would actually be minimal.

schungx (changes requested):

src/types/fn_ptr.rs
```diff
@@ -571,6 +602,33 @@ impl FnPtr {
...
+    /// `ErrorFunctionNotFound`.
+    ///
+    /// The second half is not decoration. A script body binds `this` whether or
+    /// not it reads it, so Rhai's own pointers answer `true` unconditionally. A
+    /// compiled chunk is reached through a wrapper registered at one arity, and
+    /// only a chunk that reads the receiver is reachable that way at all
+    /// (`grain/vm/callback.rs`) — so handing one to a chunk that does not take
+    /// it means calling a wrapper with an argument too many.
+    #[must_use]
+    fn declared_shape(&self) -> Option<(usize, bool)> {
```
schungx: Perhaps this should be declared as `impl FnPtrType`?

Then it is `self.typ.declared_shape()`....

schungx (changes requested):

src/types/fn_ptr.rs
schungx: In fact, I'd suggest removing this altogether when we move the logic inside `call_raw`, which simplifies the whole thing.

src/types/fn_ptr.rs
```diff
@@ -26,6 +26,32 @@ pub enum FnPtrType {
...
+    /// [`FnPtr::declared_shape`] and its one caller read. Everywhere else — the
+    /// dispatch arms in `func/call.rs`, and [`FnPtr::call_raw`] — it deliberately
+    /// falls through to dispatch by name, exactly as [`Normal`][FnPtrType::Normal]
+    /// does, reaching the chunk through the wrapper the VM registered for it
+    /// (`grain/vm/callback.rs`). That is not an oversight: routing a call
+    /// straight into the chunk is a larger change, and leaving those sites alone
+    /// keeps this variant a strict improvement on the name-only pointer rather
+    /// than a second way to call one.
+    #[cfg(all(feature = "grain", not(feature = "no_function")))]
+    Compiled {
```
schungx, line 58: Do we have to point back to the shared program here?

Can we just store a `bind_this: bool` and be done with it?

schungx (changes requested):

src/types/fn_ptr.rs
schungx: If we change to a boolean, the changes further simplify to this one-liner:

```rust
FnPtrType::Compiled { bind_this: false } => this_ptr = None,
```

src/types/fn_ptr.rs
ImTheSquid: The problem is that the way the arguments are laid out with `args` and `extras`, it can be ambiguous as to what type of resolution should occur, thus the logic for `FnPtrType::Script`. This PR essentially just applies the same logic to VM-managed functions. Your solution would result in incorrect behavior because it doesn't have the knowledge of the preferred function call arrangement from the native call.

src/types/fn_ptr.rs
ImTheSquid: See my comment on the move

src/types/fn_ptr.rs
ImTheSquid: The reason it takes the program is it can potentially escape to the host, and if that happens it needs to be able to find its way back to the `Program` it came from. The reason for the index is so it can do fast resolution, it's completely free complexity-wise since it doesn't change the ABI.

src/types/fn_ptr.rs
schungx: Well, I think it actually can be done.  And we can also do away with the entire `needs_walker` and falling back to AST-walking for closure functions using the `this` pointer.

In `callback::wrappers`, right now you skip all functions using `this` for this reason, falling back to including the AST for walker evaluation.

With `FnPtrType::Compiled` you don't have to.  You can simply just register that function.  And there will no longer be the need to `needs_walker` etc.

You just register two different types of wrappers: One without `this` usage, and one with `this`.  The version with `this` usage  (say `invoke_with_this`) simply pops the first argument as the `this` pointer, then go on with the rest.

schungx: > it can be ambiguous as to what type of resolution should occur

Actually it isn't ambiguous.  There is only one type of function VM-generated, which is a script.  It can never be Rust-native.  Therefore, it _always_ would expect the `this` pointer in the zeroth position, just like `Script`.

The only difference whether Rhai would know to put that `this_ptr` in when the VM needs it, or it doesn't if the VM doesn't want it (so it won't end up with an extra argument).

That's why I'm proposing changing this to a simple `bool`.

schungx: @ImTheSquid   Actually I have a version that does exactly this:

1) Removed `needs_walker`
2) Removed `lib` from `Program` and `Parts`
3) Only has one `bind_this: bool` in `FnPtrType::Compiled`
4) No falling back to AST-walking
5) Successfully handle callbacks in the VM that uses `this`
6) Successfully handle callbacks in the VM not using `this`

I'm trying to consolidate all of this such that it doesn't matter whether the VM function uses `this`...

Do you know how I can put a PR on top of this PR?

ImTheSquid: Removing arity from the pointer type results in a significant performance hit of like 50%. That needs to stay in. I'm also working on my own fixes and I am taking some of your suggestions, give me a bit please.

schungx: Gotcha. Just doing some experimentation on my side as well. 

I don't see how the arity issue should affect performance. In fact it should simplify the situation. 

That's because a compiled chunk has no arity conflicts. You can consider them to always be arity + 1, which are the arguments plus a `this`. It doesn't matter that the `this` pointer is not used. How I do it right now is, if it doesn't need it, Rhai binds it to a dummy temp with `()` in it. 

So the arity issue is solved. The only thing remaining is to tell Rhai that this is actually a scripted functions instead of a native function. Thats what `Compiled` tells it.

schungx:
@ImTheSquid   #1144   is there as a proof of concept.

src/types/fn_ptr.rs
ImTheSquid: I mean I ran a performance benchmark and it was significantly reduced, you can run the bench too and see what results you get. Sometimes `this` is unbound iirc so you can't just assume.

schungx: Hhhhmmm....  let me run that again.  I ran it before and the difference was within error bars...

Right now in the experiment I cheated it: if `this` is unbound, then it'll be sent to a dummy with `()` as value.

Technically speaking, it should be returning with an `ErrorUnboundThis`...

schungx:
@ImTheSquid   I've been thinking a bit more on this topic.  We might have been going at in from the wrong direction.

## The Root of the Problem

The question is: what caused this call-shape problem?

It is the distinction between a scripted function and a native Rust function.  In particular, a native Rust function cannot accept the `this` pointer... which I admit is likely a major design flaw but one that was based on history.  It was originally implemented this way from the very first versions.

Due to the fact that a native Rust function cannot accept the receiver of the function call, Rhai was adapted to allow putting the receiver as the first `&mut` slot.  Which turns into the convention that `obj.foo(1,2,3)` is the same as `foo(obj, 1, 2, 3)` if `foo` is a native Rust function.

In conclusion: the root of the problem is inconsistent treatment of function call shape between scripted and native functions.

## Our Problem

Our problem arose because the VM registers a bunch of callbacks.  They are _native_ functions, but they redirect back to the VM running _scripted_ functions.  This mismatch is causing all sorts of problems, because Rhai simply doesn't know how to treat this.

This behavior muddles the hard distinction between scripted and native functions assumed all over Rhai code.

## Maybe...

Maybe what we need is something different.

Why register a _native_ function for the callback?  Why not register a _scripted_ function instead?

The only problem stopping us is that we cannot provide an AST for the scripted function.

So, what we create a new AST node type just for this?  Call it `Stmt::GrainChunk`...

## `Stmt::GrainChunk`

`Stmt::GrainChunk` is created with a `Box<dyn>` containing the wrapper.

The VM registers its callbacks as `ScriptFunc`, with the AST body being only one statement: `Stmt::GrainChung`.

If Rhai sees `Stmt::GrainChunk`, it creates an `EvalContext`, then calls a standard VM function with that wrapper, and the VM can do the rest.

This will be a surgical change, with minimal changes on both sides, and removes the muddling of the native/scripted mismatch.

schungx:
Come to think of it, it is not even necessary for a new AST node type. 

Just put a shared VM inside the `ScriptFunc` struct and hijacks it when the function is run. 

This can really be minimal change. I'm gonna try it out.

ImTheSquid:
I think having an enum on `ScriptFuncDef` is a smart way to do it, but this PR should still be merged for correctness purposes. The `ScriptFunc` just makes it way faster. We've already patched 20k+ lines into this release, 500 more to fix weird dispatching rules won't hurt.

schungx:
> but this PR should still be merged for correctness purposes

Give me a bit of time to try out my idea first.  If it works, then the whole patching to `fn_ptr.rs` won't be necessary, as with introducing a new `FnPtrType`.  So there would be a lot of changes to back out if that idea works...

And I assume this PR does not handle closures using the `this` pointer, redirecting them to the AST walker?

schungx:
> I think having an enum on `ScriptFuncDef` is a smart way to do it

Well, your excellent suggestion works out!

I changed the payload of the `ScriptFuncDef` into an enum and it just works perfectly!

And without really touching anything... None of the files are changed in any major way, except to handle the new enum.

`takes_this` is also no longer necessary as it doesn't matter whether a scripted function uses `this` or not.

So far it passes all the tests. I'll test it some more then push a branch.

ImTheSquid:
Yeah I think we should combine our implementations and merge them in a stacked PR. Let me know how all of that goes, would love to have this all figured out

schungx:
> Yeah I think we should combine our implementations and merge them in a stacked PR. Let me know how all of that goes, would love to have this all figured out

Ok, here we go again. I've uploaded #1153 which is a different approach. I think it is better because we're not trying to get around Rhai. We'll be changing Rhai to support Grain.

I wonder how I can merge two PRs... Just cherry pick?

ImTheSquid:
I think we should take the tests from my branch and put it into yours. Then we can see if yours passes all the tests. I think #1153 is the best approach so far if the tests all pass

schungx:
> I think we should take the tests from my branch and put it into yours. 

Yeah, I'll do that in a bit. Also I want to make one small change to cache in the chunk number instead of always going through a name search for all calls... Name searches in the VM are expensive as we need to keep decoding the bytestream.

BTW... As I mentioned there, there is this `vm.adopt` business that I'm not sure what it is... Something to keep the trace or context across invocations...  So pls have a look at that! 😅

ImTheSquid:
Closed in favor of #1153

PR #1147, opened 2026-08-20
VM: Implement indexer fallback for missing properties.

schungx:
Right now, the VM does not fallback to trying an indexer (getter or setter) when access to a property failed.

This PR implements this behavior so it'll match the AST walker.


ImTheSquid (changes requested):
Unused variable warnings should be handled through this kinda deal:
```rs
fn foo(
#[cfg_attr(feature="no_*", allow(unused))] var: Typ
) {}
```
Other than that LGTM

schungx:
> Unused variable warnings should be handled through this kinda deal:

Are there unused?  Let me double check...  I thought I caught them all...

EDIT: Ah, gotcha.

schungx:
> Unused variable warnings should be handled through this kinda deal:

It was too many attributes to put on the parameters, I ended up using this:

```rust
let _ = (target, key, value, fail_silently, pos);
```

This will get removed by the compiler during optimization so it won't affect performance.

I see a couple of places in `grain` that this style is used, so I suppose this is OK!  😆

ImTheSquid (approved):

PR #1148, opened 2026-08-21
VM: (EXPERIMENTAL) Change to a capabilities-based system from feature flags.

schungx:
This PR is an experiment to replace the current feature flags-based fingerprint for artifacts with one based on the _caps_ supported by a particular host build, as well as the caps required to run an artifact.

## `Caps`

The `Caps` bit-flags contains a bunch of capabilities for running an artifact inside the VM.

## Host Caps

Based on the build's feature flags, a set of caps for the host VM can be obtained (together with other info such as the width of integers and floats etc.).

## Compilation

During compilation, when the compiler lowers from AST nodes, the caps are collected.  In the end, there is a set of caps required for running that artifact, which is stored as a fingerprint together with the artifact.

## Verification

The verifier supports caps as well.  When an artifact is being verified, it goes through all the instructions and checks that they never require caps not available in the supported set.

## Reading

Reading from a stored artifact compares the required caps (stored as fingerprint) with what the host supports.  This is quite versatile.

For example, assume that an artifact is built without `no_index`, but the script does not use any arrays or indexing.  Then the artifact would not require the `INDEXING` or `ARRAY` caps.  Such an artifact can successfully run on a VM built with `no_index`.  That's because, even though the host caps do not contain `INDEXING` nor `ARRAY`, they are not needed to run the artifact and therefore the load is allowed to succeed.

## Safeguard

When the artifact is loaded, the verifying process checks the actual instruction caps with the stored fingerprint to make sure that nothing has been mucked with.  That is to safeguard the situation that the fingerprint is manually edited afterwards to drop a necessary cap, which may result in errors when it is run on the VM.


src/grain/format/abi.rs
```diff
@@ -19,31 +19,142 @@
...
+bitflags! {
+    /// Capability flags.
+    ///
+    /// Order is the wire order and must never change; append only.
+    #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct Caps: u32 {
+        /// The script uses floating-point numbers, which are not available under `no_float`.
+        const FLOAT = 0b0000_0000_0000_0000_0001;
+        /// The script uses arrays, which are not available under `no_index`.
+        const ARRAY = 0b0000_0000_0000_0000_0010;
```
ImTheSquid: Bit of a nit but I think it's better to use `1 << N` instead of writing out the entire binary number, it's hard to find the one here

src/grain/format/abi.rs
```diff
@@ -19,31 +19,142 @@
...
+        const IMPORT = 0b0000_0000_1000_0000_0000;
+        /// The script uses the `export` statement to export in modules, which is not available under `no_module`.
+        const EXPORT = 0b0000_0001_0000_0000_0000;
+        /// The script uses the custom syntax, which is not available under `no_custom_syntax`.
+        const CUSTOM_SYNTAX = 0b0000_0010_0000_0000_0000;
+
+        /// The script is built with `sync`.
+        const SYNC = 0b0100_0000_0000_0000_0000;
+        /// The script contains unsupported syntax.
+        const UNSUPPORTED = 0b1000_0000_0000_0000_0000;
```
ImTheSquid: I don't really see why this is necessary, the program can't be exported if it has unsupported items so this will never be written.

src/grain/format/abi.rs
```diff
@@ -19,31 +19,142 @@
...
+        const SHARING = 0b0000_0000_0100_0000_0000;
+        /// The script uses the `import` statement to import modules, which is not available under `no_module`.
+        const IMPORT = 0b0000_0000_1000_0000_0000;
+        /// The script uses the `export` statement to export in modules, which is not available under `no_module`.
+        const EXPORT = 0b0000_0001_0000_0000_0000;
+        /// The script uses the custom syntax, which is not available under `no_custom_syntax`.
+        const CUSTOM_SYNTAX = 0b0000_0010_0000_0000_0000;
+
+        /// The script is built with `sync`.
+        const SYNC = 0b0100_0000_0000_0000_0000;
```
ImTheSquid: This is more of a question, but what does `sync` change here? What capabilities does that restrict inside the binary?

ImTheSquid (changes requested):

src/grain/format/abi.rs
schungx: Yeah, really. That's a good point.

src/grain/format/abi.rs
schungx: True...

src/grain/format/abi.rs
schungx: Actually nothing really... Probably should be removed.

schungx:
@ImTheSquid  I'm unsure how to proceed with this. 

Do we release 1.26 first? In that case the fingerprint will be feature flag based. When we (or if we) change it later on, it'd be a breaking change for all users with saved artifacts. 

Do we put this in for 1.26? In that case there is probably not enough time to fully test it out... 

Although the capabilities concept is probably more resilient and flexible than feature flags, but feature flags are unlikely to change so they're stable!

ImTheSquid:
I think this can merge if we add some more tests. I like the idea of this a lot and this is definitely the best iteration of it so far. If we need to change it after 1.26 that's fine, Grain is still gonna be experimental for a version or two and we will make the ABI versioning clear.

src/grain/format/abi.rs
```diff
@@ -1,49 +1,129 @@
...
+    /// Capability flags.
+    ///
+    /// Order is the wire order and must never change; append only.
+    #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct Caps: u32 {
+        /// The script uses floating-point numbers, which are not available under `no_float`.
+        const FLOAT = 1<<0;
+        /// The script uses arrays, which are not available under `no_index`.
+        const ARRAY = 1<<1;
+        /// The script uses BLOB's, which are not available under `no_index`.
```
ImTheSquid, line 17: There are a few duplicates here in terms of mapping a feature flag to a `Caps`. Is this intentional? I don't really see a case where an array would be allowed but a blob wouldn't and vice versa since they're both arrays under the hood.

schungx: Yes, I was trying to break it down as finely as reasonable. You're right that some of them go together, but not necessarily will always be the case.

We can change it to reflect mostly the feature flags. That won't affect things.

schungx: One other reason for breaking it down into smaller units: we can probably use those capabilities for other purposes, such as to analyze a script. 

I suggest we keep this fine-grained caps just in case future uses come up.

schungx:
@ImTheSquid   Good thing you asked me to add more tests.

The new tests flushed out a bug: I missed the `DEFINE_FUNCTION` cap.

ImTheSquid (approved):

PR #1149, opened 2026-08-22
VM: Reuse shadowed variable slots

schungx:
A minor refinement.

Rhai tries to reuse a scope slot if the variable has been shadowed by a new one with the same name.

If shadowed within the same block, it is impossible to access the old variable again, so its slot can be reused.  This is Rhai's behavior.

This PR adds versions of opcode that declares a variable into an existing slot instead of always putting it on top of the scope.

### DIFF
--- a/examples/grain_bench.rs
+++ b/examples/grain_bench.rs
@@ -78,7 +78,7 @@ const CASES: &[Case] = &[
     },
     Case {
         name: "script fn calls",
-        source: "fn add(a, b) { a + b } let s = 0; for i in 0..5000 { s = add(s, i); i += 1; } s",
+        source: "fn add(a, b) { a + b } let s = 0; for i in 0..5000 { s = add(s, i); } s",
         iterations: 20,
         callbacks: false,
         floor: 1.55,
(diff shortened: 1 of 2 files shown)


src/grain/bytecode/code.rs
```diff
@@ -179,6 +179,10 @@ pub mod tag {
     pub const CALL_THIS_REF_CAPTURE: u8 = 0x46;
     /// [`Op::Statement`](super::Op::Statement).
     pub const STATEMENT: u8 = 0x47;
+    /// [`Op::DeclareLocal`](super::Op::DeclareLocal) into a specific `slot`.
+    pub const DECLARE_LOCAL_SLOT: u8 = 0x48;
```
ImTheSquid: why does this need a new opcode? can't we just use assignment internally?

ImTheSquid (changes requested):

src/grain/bytecode/code.rs
schungx: That's a good point. Not sure why not... Technically speaking it can work with an assignment... 

I vibecoded this so this was what copilot suggested. Maybe AI is not so bright after all... 

How do I change this to an assignment? 

In that case the `slot` needs not even be in`DeclareLocal`...

schungx:
OK, making it into a store solves the entire problem!

Some condolences... AI is not replacing us humans _yet_...

ImTheSquid (approved):
Beautifully short

schungx:
> Beautifully short

A good feeling... AI doesn't know everything... Yet.

PR #1153, opened 2026-08-23
VM: Implement callbacks from native Rust code

schungx:
This is a re-architect of the attempt to make Grain function chunks callable from native functions registered in Rhai in a seamless manner that all existing code would work.

This approach starts from a different direction -- instead of fitting Grain into an unaccommodating Rhai, Rhai is taught that different types of scripted functions exist.

Therefore, it replaces a scripted function's body (normally a `StmtBlock`) with an enum specifying the type of the payload.

* A `Statements` payload works the same as before.

* A `GrainVM` payload is given a shared `Program` to run.

Thus nothing changes in Rhai, nor in Grain.  You can observe that none of the files touched by previous attempts (e.g. `fn_ptr.rs`) is modified.

Rhai now knows that a Grain wrapper is a scripted function (not disguised as a native) and calls it directly.

A side benefit is that `takes_this` is no longer necessary. There is no difference in call shapes as Rhai knows the function is a script.

@ImTheSquid   I'm quite sure this is missing the `vm.adopt(program)` and `release` etc. that has to do with multiple levels of reentrency...

### DIFF
--- a/src/ast/script_fn.rs
+++ b/src/ast/script_fn.rs
@@ -31,7 +31,8 @@ impl ScriptFuncPayload {
     pub const fn position(&self) -> Position {
         match self {
             Self::Statements(block) => block.position(),
-            _ => Position::NONE,
+            #[cfg(feature = "grain")]
+            ScriptFuncPayload::GrainVM(..) => Position::NONE,
         }
     }
     /// Get the end position.
@@ -40,7 +41,8 @@ impl ScriptFuncPayload {
     pub const fn end_position(&self) -> Position {
         match self {
             Self::Statements(block) => block.end_position(),
-            _ => Position::NONE,
+            #[cfg(feature = "grain")]
+            ScriptFuncPayload::GrainVM(..) => Position::NONE,
         }
     }
 }
@@ -55,9 +57,9 @@ pub struct ScriptFuncDef {
     pub name: ImmutableString,
     /// Function access mode.
     pub access: FnAccess,
-    #[cfg(not(feature = "no_object"))]
     /// Type of `this` pointer, if any.
     /// Not available under `no_object`.
+    #[cfg(not(feature = "no_object"))]
     pub this_type: Option<ImmutableString>,
     /// Names of function parameters.
     pub params: FnArgsVec<ImmutableString>,
--- a/src/ast/script_fn.rs
+++ b/src/ast/script_fn.rs
@@ -2,17 +2,50 @@
 #![cfg(not(feature = "no_function"))]
 
 use super::{FnAccess, StmtBlock};
-use crate::{FnArgsVec, ImmutableString};
+use crate::{FnArgsVec, ImmutableString, Position};
 #[cfg(feature = "no_std")]
 use std::prelude::v1::*;
 use std::{fmt, hash::Hash};
+
+/// A type containing the body of a script-defined function.
+#[derive(Debug, Clone)]
+pub enum ScriptFuncPayload {
+    /// Normal statements block.
+    Statements(StmtBlock),
+}
+
+impl Default for ScriptFuncPayload {
+    #[inline(always)]
+    fn default() -> Self {
+        Self::Statements(StmtBlock::NONE)
+    }
+}
+
+impl ScriptFuncPayload {
+    /// Get the start position.
+    #[inline(always)]
+    #[must_use]
+    pub const fn position(&self) -> Position {
+        match self {
+            Self::Statements(block) => block.position(),
+        }
+    }
+    /// Get the end position.
+    #[inline(always)]
+    #[must_use]
+    pub const fn end_position(&self) -> Position {
+        match self {
+            Self::Statements(block) => block.end_position(),
+        }
+    }
+}
 
 /// _(internals)_ A type containing information on a script-defined function.
 /// Exported under the `internals` feature only.
 #[derive(Debug, Clone)]
 pub struct ScriptFuncDef {
     /// Function body.
-    pub body: StmtBlock,
+    pub body: ScriptFuncPayload,
     /// Function name.
     pub name: ImmutableString,
     /// Function access mode.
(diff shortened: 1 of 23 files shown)


schungx:
@ImTheSquid  I've added the tests from your PR and more of my own stressing the use of `this`).

So far they all pass. ✌️

src/optimizer.rs
```diff
@@ -1513,11 +1513,27 @@ impl Engine {
...
             crate::Module::from(functions.into_iter().map(|fn_def| {
                 // Optimize the function body
-                let mut fn_def = crate::func::shared_take_or_clone(fn_def);
-                let statements = fn_def.body.take_statements();
-                *fn_def.body.statements_mut() =
-                    self.optimize_top_level(statements, scope, lib2, optimization_level);
-                fn_def.into()
+                match fn_def.body {
+                    // Payload is a statements block
+                    crate::ast::script_fn::ScriptFuncPayload::Statements(..) => {
```
ImTheSquid, line 1518:
```suggestion
                    crate::ast::script_fn::ScriptFuncPayload::Statements(body) => {
```

Why not just extract `body` here?

src/grain/vm/mod.rs
```diff
@@ -386,7 +386,7 @@ pub struct Vm<'e> {
     /// `this` never inherited: a plain nested call passes `None` and gets the
     /// caller's back on the way out, reproducing `func/call.rs:669` without a
     /// conditional anywhere.
-    this: Option<Dynamic>,
+    this_ptr: Option<Dynamic>,
```
ImTheSquid: Why rename? Kinda pollutes the PR

ImTheSquid (changes requested):
I like where this is going, just a few small things

src/grain/vm/mod.rs
schungx: Well, to be consistent with other parts of Rhai. Nothing very special. Will revert. 😅

ImTheSquid:
Also failing test ofc

src/optimizer.rs
schungx: That's because of this line
```rusr
let mut fn_def = crate::func::shared_take_or_clone(fn_def);
```

It consumes the `fn_def` within an `Rc` (if it isn't shared, mostly not when in optimization step). Consumes it, change it, then write back the optimized version. 

And the `match` afterwards with the `unreachable!()` for the Grain path is just to shut up clippy warnings for different feature flag combos.

schungx:
> Also failing test ofc

Yes, the error structure is different... I'll take a look at it. 

EDIT: Yes, it is a Rhai bug. Once again th VM has flushed out bugs.

src/optimizer.rs
ImTheSquid: Ah I see, makes sense. Maybe there's a mode idiomatic way of doing this but maybe just put a comment there explaining why you do the second `match`

schungx:
Rebase on latest.  Pls give it a final round of looks...

schungx:
Since it is related, I just put in the implementation for OOP-style calling.  Not difficult at all, just an extra special condition check.

Now I believe 99% of Rhai's tests suite passes.

Soooo....  this is it.  The last big PR, after which it's 1.26.

src/ast/script_fn.rs
```diff
@@ -2,24 +2,68 @@
...
+
+impl ScriptFuncPayload {
+    /// Get the start position.
+    #[inline(always)]
+    #[must_use]
+    pub const fn position(&self) -> Position {
+        match self {
+            Self::Statements(block) => block.position(),
+            #[cfg(feature = "grain")]
+            ScriptFuncPayload::GrainVM { .. } => Position::NONE,
```
ImTheSquid: this isnt necessarily true, the line information is there if the program isn't stripped

src/func/call.rs
```diff
@@ -728,8 +728,13 @@ impl Engine {
...
 
-                if fn_def.body.is_empty() {
-                    return Ok((Dynamic::UNIT, false));
+                match fn_def.body {
+                    crate::ast::script_fn::ScriptFuncPayload::Statements(ref block)
+                        if block.is_empty() =>
+                    {
+                        return Ok((Dynamic::UNIT, false))
+                    }
+                    _ => (),
```
ImTheSquid: not a fan of the catchall

src/func/script.rs
```diff
@@ -43,13 +43,24 @@ impl Engine {
...
+            ScriptFuncPayload::Statements(ref block)
+                if self.debugger_interface.is_none() && block.is_empty() =>
+            {
+                return Ok(Dynamic::UNIT)
+            }
+            #[cfg(not(feature = "debugging"))]
+            ScriptFuncPayload::Statements(ref block) if block.is_empty() => {
+                return Ok(Dynamic::UNIT)
+            }
+            _ => (),
```
ImTheSquid: same as above

src/grain/compile/mod.rs
```diff
@@ -870,17 +863,24 @@ impl Lowering {
             .map(|p| self.push_name(p.clone()))
             .collect();
 
+        let body = match &def.body {
+            // The scripted function's body must be an AST statements list.
+            ScriptFuncPayload::Statements(body) => body,
+            // For some reason, if it is a Grain chunk, then fail to lower.
+            ScriptFuncPayload::GrainVM { .. } => return None,
```
ImTheSquid: is this even possible? this seems like the wrong path here. if its somehow from a *different* program then couldn't we add it into this new one being compiled?

ImTheSquid (changes requested):

src/ast/script_fn.rs
schungx: Hhmmm... let me see how I can put the position in.

src/grain/compile/mod.rs
schungx: This is not really possible, as a `GrainVM` function can ever only be handed over by Grain inside a callback wrapper.

Therefore it returns `None` as being unable to lower.  I assume this is better then using `unreachable!()` or `expect` that will panic when reached.

Or what do you suggest we do in this case?

ImTheSquid: if its not possible then we should put an `unreachable!()` here

ImTheSquid (approved):

PR #1156, opened 2026-08-26
Optimize out Unit/Pop pairs, control flow structure jumps

ImTheSquid:


schungx (approved):
I wonder how much speed up we can get .. 

It certainly would have reduced the payload a bit for sure...

schungx (approved):

schungx:
I think this is ready to merge, unless you have more changes.

PR #1159, opened 2026-08-27
VM: Improve callback performance 

schungx:
This is it.

### Simplified Rhai `FnPtr` data structure

Due to [this commit](https://github.com/rhaiscript/rhai/commit/648c684861ad143067b67c4e444b389ddb9f1015), Rhai's `FnPtr` type no longer contains any _embedded environment_, nor does it contain any pointers to internal data structures.

In hindsight, it was a poor design decision anyway. I was too eager to make something work.

In other words, the `FnPtr` type is now completely serializable and can be considered a normal type with native support by the VM.

### What this PR does

Add support for loading `FnPtr` directly from a `Expr::DynamicConstant`.

Previous this was farmed out into the `MakeClosure` op.  The reader/writer now considers it a known type serialized/deserialized natively.

### Impacts

1) The `MakeClosure` op is no longer used (but kept for now).

2) There is no longer _any_ distinction between a `FnPtr` created by the walker and a `FnPtr` created by the VM.

3) There is no more speed penalty in closure callbacks into the VM.  Benchmarks now show callbacks to be at least at par with the walker.

4) Large sections of code in Rhai (and some in the VM) became unnecessary and removed, simplifying the codebase significantly.

5) All the special tests for differences between `FnPtr` now return them as identical. All tests related to `FnPtr` that used to fail now pass and patched accordingly. Docs related to this difference are also patched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Version 1.27.0
 Bug fixes
 ---------
 
-* Fixed bug in using bare function names as a variable (identified by Grain VM, [`#1158`](https://github.com/rhaiscript/rhai/pull/1158)).
+* Fixed bug in using bare function name as a variable (identified by Grain VM).
 
 
 Version 1.26.0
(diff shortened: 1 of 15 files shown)


ImTheSquid (approved):
So cool!!! Thank you for pushing all of this stuff out, I've been very busy lately but good to see you're finding so many ways to improve Grain.

Issue #1098, opened 2026-05-27
Version 1.25.0 had undocumented breaking change


Abendrill:
> This removal was a breaking change as dependent crates could import these items via `rhai::plugin`. For example [apollo-router](https://github.com/apollographql/router/blob/dev/apollo-router/src/plugins/rhai/engine/mod.rs) are (and have for quite some time) been relying on this import path. 

 _Originally posted by @Abendrill in [475d45e](https://github.com/rhaiscript/rhai/commit/475d45ee294aa4cf95c7e07bdca2934dca10e20b#r186669870)_

Is this an unintended breaking change? If so, could it be re-added?
Or if it is, could that be reflected in the version?

----

**Note:** I'm doing a custom build with Rust plugins. Therefore, because `apollo-router` is used as a library, your `Cargo.lock` file [is ignored](https://doc.rust-lang.org/1.71.1/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries). Therefore, `rhai` is resolved to anything that should be semver compatible with the version you've specified (`1.23.6`). That's why cargo by default resolves `rhai@1.25.0` for me and not `rhai@1.23.6`.

schungx:
Yes I can see the problem now. 

The exports should be added back in and kept until the next major version. 

Anyone care to make a PR? I can make one if no one does.

Abendrill:
PR posted ⬆️

PR #1160, opened 2026-08-31
Add no_ast feature.

schungx:
This is it.  The `no_ast` feature, that excludes:

- The `AST`
- All `AST` interpretor
- The parser
- The tokenizer
- The optimizer

Right now, it must be used with `grain`, otherwise this would be a quite boring build of Rhai -- doing nothing.  And yes, it _can_ be built with just `no_ast` and no `grain`.

`no_ast` automatically implies `no_optimize` (obviously), as well as `no_module`, `no_time` and `no_custom_syntax` (because Grain doesn't support it yet).  It also cannot be used with `debugging`.

Under `no_ast`, Grain excludes the Compiler, `residuals` as well as the `EvalAst` opcode.  It is a testimonial towards the well-structured design of Grain that it took minimal modifications to accommodate the `no_ast` feature.  Essentially, it was little more than just gating the `compiler` module.

I would be _really_ interested to know how small a binary this can become...  when only bytecodes are being executed.


.github/workflows/build.yml
```diff
@@ -236,7 +258,7 @@ jobs:
           RUSTDOCFLAGS: "-D warnings"
         with:
           command: doc
-          args: --no-deps --features document-features,metadata,serde,internals,decimal,debugging
+          args: --no-deps --features document-features,metadata,serde,internals,decimal,debugging,grain
```
ImTheSquid, line 263: This job now fails. Adding `grain` surfaces a third rustdoc error that the PR doesn't fix:

```
error: redundant explicit link target
  --> src/lib.rs:1:1
  explicit target `grain/index.html` is redundant because label `grain` resolves to same destination
```

Source is `Cargo.toml:81` (`## Enable the experimental [`grain`](grain/index.html) bytecode VM`), pulled in by `document-features` — with `grain` on, `` `grain` `` resolves by itself. The two `crate::grain::restore::restore` errors are fixed here; this one isn't. Suggest a non-self-resolving label, e.g. `[Rhai Grain](grain/index.html)`.

src/api/events.rs
```diff
@@ -56,6 +58,7 @@ impl Engine {
     /// # }
     /// ```
     #[deprecated = "This API is NOT deprecated, but it is considered volatile and may change in the future."]
+    #[cfg(not(feature = "no_ast"))]
```
ImTheSquid: The Grain VM reads `engine.resolve_var` in five places (`grain/vm/mod.rs:1244, 1997, 2048, 2097, 4153`), but `on_var` is the only way to set it — so under `no_ast` the field is permanently `None` and those paths are dead. Note the field itself isn't gated (unlike `token_mapper`, which correctly is).

```
error[E0599]: no method named `on_var` found for struct `Engine`   // features = no_ast, grain
```

`OnVarCallback` is `dyn Fn(&str, usize, EvalContext) -> RhaiResultOf<Option<Dynamic>>` — nothing AST-specific. Gate should come off. (`on_def_var` is fine: `def_var_filter` is only read from `parser.rs` and `eval/stmt.rs`.)

Same class, lower severity: `on_invalid_array_index` / `on_map_missing_property` are gated off, but their call sites in `get_indexed_mut` are reached by the VM (`grain/vm/mod.rs:1530`). Gated consistently so nothing is dead, but a `no_ast+grain+internals` build loses them for no reason.

src/func/plugin.rs
```diff
@@ -14,8 +14,8 @@ pub use rhai_codegen::*;
 // https://github.com/rhaiscript/rhai/issues/1098
 pub use super::RhaiFunc;
 pub use crate::{
-    Dynamic, Engine, EvalAltResult, FnAccess, FnNamespace, FuncRegistration, ImmutableString,
-    Module, NativeCallContext, Position,
+    Dynamic, Engine, EvalAltResult, FnNamespace, FuncRegistration, ImmutableString, Module,
+    NativeCallContext, Position,
```
ImTheSquid: `FnAccess` dropped from this list — breaking change, and the comment three lines above says exactly that (#1098). Confirmed `rhai::plugin::FnAccess` is now `E0425`.

Nothing forces it: `crate::FnAccess` still resolves, it just comes from `func` instead of `ast` now. Re-adding it here is a one-word fix.

src/lib.rs
```diff
@@ -276,13 +287,17 @@ pub use func::Locked;
...
 #[cfg(not(feature = "no_function"))]
+#[cfg(not(feature = "no_ast"))]
 pub use func::Func;
 
+#[cfg(feature = "internals")]
+pub use func::ScriptFuncDef;
+
 #[cfg(not(feature = "no_function"))]
-pub use ast::ScriptFnMetadata;
+pub use func::{EncapsulatedEnviron, ScriptFnMetadata};
```
ImTheSquid: `EncapsulatedEnviron` was `internals`-only on main (`#[cfg(feature = "internals")] pub use ast::{… EncapsulatedEnviron …}`). It now compiles in a default build with no `internals` — verified. Probably wants its own `#[cfg(feature = "internals")]` line.

src/eval/target.rs
```diff
@@ -42,12 +42,12 @@ pub fn calc_offset_len(length: usize, start: crate::INT, len: crate::INT) -> (us
 /// Values going over bounds call the provided closure to return a default value or an error.
 #[inline]
 #[allow(dead_code)]
-pub fn calc_index<E>(
+pub fn calc_index(
```
ImTheSquid, line 45: Any reason to drop the `<E>`? It forces `array_basic::remove` and `blob_basic::remove` to build a throwaway boxed `ErrorRuntime(Dynamic::UNIT, Position::NONE)` where they used `Err(())` as a free bail sentinel. Same behavior, but an allocation on the out-of-bounds path and it reads oddly.

.github/workflows/build.yml
```diff
@@ -156,6 +156,28 @@ jobs:
           command: build
           args: ${{matrix.flags}}
 
+  no-ast:
+    name: Check no_ast build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        flags:
+          - "--features no_ast,decimal,metadata,grain"
```
ImTheSquid, line 166: Thin coverage. The description says `no_ast` alone works — it builds, but emits 11 warnings, including 2 `unreachable statement` (`func_call.rs:849`, `script.rs:331`, from the uninhabited `ScriptFuncPayload`). Worth adding `--features no_ast` and `--features no_ast,internals,sync,grain`.

Cargo.toml
```diff
@@ -102,6 +102,8 @@ no_closure = []
 no_module = []
 ## Remove support for custom syntax.
 no_custom_syntax = []
+## Remove the AST, tokenizer, parser and optimizer.
+no_ast = ["no_module", "no_custom_syntax", "no_optimize", "no_time"]
```
ImTheSquid: `no_time` is a Grain limitation, not an AST one — encoding it here means a downstream crate enabling `no_ast` silently forces `no_time` on everyone in the graph (features are additive), and relaxing it later is awkward. Same argument applies to `no_module`.

Cargo.toml
```diff
@@ -132,14 +134,15 @@ testing-environ = []
 
 [[bin]]
 name = "rhai-repl"
-required-features = ["rustyline"]
+required-features = ["bin-features"]
 
 [[bin]]
 name = "rhai-run"
+required-features = ["bin-features"]
```
ImTheSquid: `cargo run --bin rhai-run script.rhai` now fails without `--features bin-features`. Understandable as the mechanism for keeping bins out of `no_ast` builds (via `debugging`), but it's an unannounced UX change — worth a CHANGELOG line.

ImTheSquid (commented):
Checked the moves line-by-line (sorted line-set diffs of old vs. new files) — `tokenizer.rs` → `types/token.rs`, `chaining.rs` → `indexing.rs`, `call.rs` → `func_call.rs`/`script.rs`/`function.rs`, and the `ast/*` → `func/*` moves are all content-preserving. Tests pass across the CI matrix, `cargo fmt --check` is clean, and all the `no_ast` combos I tried build (`no_ast` alone, `+grain`, `+internals`, `+sync`, `+no_function`, `+no_object,no_index`, `+unchecked`).

A few real issues below — one breaks CI, two are unintended public-API changes.

ImTheSquid:
Most of this PR is good and very exciting, just need to make sure we do it right. The diff did lead me to do an AI review, forgive me lol

schungx:
Great you found all these issues! Your AI seems better than mine... Which one do you use? 

I tried doing the `no_ast` feature with Copilot which primarily uses GPT 5.3 (I didn't pay for the Pro+ tier).

It got stuck in a loop going nowhere and just consumed large amounts of tokens doing nothing. In the end I just spent about 4 hours over the weekend to manually rearchitect it. 

I have been hoping that AI can't replace us... yet...

ImTheSquid:
lol this was all Opus 5 med thinking, it's my main workhorse

.github/workflows/build.yml
schungx: These unused/unreachable warnings are expected.  `no_ast` is such a deep surgery that I have not been trying to get rid of all warnings.  The unused/unreachable code would be eliminated by Rust in release builds anyway.

schungx: Will add the new tests

Cargo.toml
schungx: Ah, yes.  `rhai-run` is a CLI tool.  It doesn't really need `bin-features`.

However, I can't see any way to avoid it being built when `no_ast` is turned on...  

Instead, maybe I should extend it such that it _also_ runs Grain bytecodes?  Now that seems like an idea!

schungx: `rhai-run` now also supports running Grain bytecodes!

If `no_ast` is on, but no `grain`, `rhai-run` still builds but does nothing!

schungx:
I benchmarked the `rhai-run` build sizes (on x64 Windows, release):

| Build | Size (MB) |
|:--|--:|
| AST + Grain | 3.43MB |
| AST only | 3.16MB |
| Grain VM only | 2.43MB |
| None | 133KB |

Linux sizes should be in similar ratios.  Of course, this is an x64 system, so it is 64-bit code.  Compiling to 32-bit with `only_i32` and `f32_float` would make things much smaller...

Which means, rough guesstimates (three equations, four unknowns):

| Module | Size (KB) |
|:--|--:|
| Tokenizer, parser, optimizer, AST walker |  860KB |
| Grain Compiler | 170KB |
| Grain VM | 100KB |
| `Engine` + libraries |  2.17MB |
| Base | 133KB |

So the bulk is still in the `Engine` infrastructure, i.e. to deal with registering and calling functions, plus the standard library.

schungx:
Also, while at it, `grain_dump` is now split into `grain-compile` and `grain-dump`, one produces the `.rgrn` file and one dumps it.

They are such nice tools that I'm moving them into `bin`.

ImTheSquid (approved):

schungx:
Updated figures:

| Build | Size (MB) |
|:--|--:|
| AST + Grain | 3.43MB |
| AST only | 3.16MB |
| Grain VM only | 2.43MB |
| Grain VM only (no libraries)| 890KB |
| None | 133KB |

| Module | Size (KB) |
|:--|--:|
| Tokenizer, parser, optimizer, AST walker |  860KB |
| Grain Compiler | 170KB |
| Grain VM | 100KB |
| Libraries | 1.28MB |
| `Engine` |  790KB |
| Base | 133KB |

ImTheSquid:
Wow grain VM is wayyyy smaller than I thought, excited for this to merge!

schungx:
> Wow grain VM is wayyyy smaller than I thought, excited for this to merge!

Merging it now.

But it seems the standard libraries are huge...  probably not small for embedded.  I suppose a couple MB's are gonna be big for the ESP32...

ImTheSquid:
If we could make the std libraries smaller that would be the big next step, but one thing at a time

PR #1165, opened 2026-09-09
VM: Double switch statement speed by not emitting the ranges dispatch if there are no range cases

schungx:
A very simple optimization for the `switch` statement.

Currently, a single `switch` statement lowers into two calls to `Op::Switch` -- one for the cases and one for the ranges.

In the predominant scenario of no range cases, there is no need to emit the dispatch to empty ranges.

This PR optimizes the lowering such that no second dispatch is emitted if there are no range cases.  It also saves the push and pop of the `Scope` by not storing the `switch` value.

Speed of the `switch` statement doubles to close to 2x the speed of the AST walker when there are no range cases.


src/grain/compile/mod.rs
```diff
@@ -562,23 +562,29 @@ impl Lowering {
...
 
-        // Store the subject's value because if all the arms decline,
-        // the ranges still needs it.
-        self.emit(Op::DeclareLocal {
-            name: value_name_index,
-            is_const: false,
-        });
-        self.emit(Op::LoadLocal(value_slot));
+        let value_slot = if !ranges.is_empty() {
+            let value_name = ImmutableString::from("$SWITCH_VALUE$");
```
ImTheSquid: Are we sure that hardcoding this is safe? This should probably be a mangled string

schungx: Anything put inside the scope is ultimately leakable... I'll see if I can put it on the stack instead...

schungx:
@ImTheSquid   I have re-architect the PR.  The push to `scope` is removed completely.  The `scope` is no longer touched.

Instead, the `Switch` op is modified so that it does not pop the value, only inspects it.  This is done to allow a subsequent `Switch` op call to also inspect eh value.

The compiler only emits a pop at the end of the `switch` statement running to completion.

The only tricky aspect is that those "stack surpluses" must also be popped if there is a `break` statement terminating an outer loop prematurely. So I keep a count of how many of these surpluses exist.

ImTheSquid (approved):

PR #1169, opened 2026-09-13
VM: Optimize loops by not emitting a `()` if not needed.

schungx:
Currently, any loops (e.g. `for`, `while`, do`) emit a `()` with a `Pop` in the vast majority of cases where the loop's break value is not used.

This results with a push `()` followed by `Pop` for many loops.  Although most loops are long enough to make this immaterial, is it nevertheless not necessary.

This PR optimizes this with the `Lowered` return value just like for statement blocks.


ImTheSquid (approved):

schungx:
There are still CI failing. I'll fix them before merging.

PR #1173, opened 2026-09-21
VM: Support Custom Syntax

schungx:
# Introduction

Custom syntax is a unique feature of Rhai, one used by many domain-specific applications to simplify code by providing new syntax suited to the domain usage.

# Grain Support
Previously, Grain rejects all custom syntax as fragments.  This PR adds support for custom syntax that does not touch the `scope`.

Custom syntax that touches the `scope`, like `eval`, remain non-lowerable. These cases are extremely rare.

# Implementation

Grain implements custom syntax via a new Opcode `Op::CustomSyntax` and a `custom_syntax` pool.

# `no_ast`

Finally, `no_ast` works with custom syntax, disabling only the parser half of it, but allows the execution part to work with Grain.


schungx:
@ImTheSquid   Finally got all the CI tests to pass and all the feature combinations properly gated!

Pls take a read thru this...  This look like a larger PR but really not so.  One new opcode, one new pool, and the relevant implementation.

All other file changes are properly gating the `no_ast` feature, which used to automatically turn on `no_custom_syntax` but now needs to co-exist with custom syntax (because Grain can run it).

